PR #140, opened 2023-06-20
fix: Adding Advertising struct to extn response. fixes #139

### LINKED ISSUE #139: Advertising response struct support to extn response.

kvfasil:


satlead:
Create a new PR with Email and CLA

PR #153, opened 2023-07-06
fix: Usergrants support for Pin Challenge

Karthick-Somasundaresan:


satlead:
Probably change the name of the commit to 

fix: Usegrants support for pin challenge

PR #176, opened 2023-07-25
feat: Certification Updates + Use Workspace version across packages

satlead:
- [x] Add Workspace version across packages
- [x] Fix Authorized Info processor to return accurate permissions
- [x] Remove fail open check for Lifecycle management
- [x] Add response when a processor is not available

### DIFF
--- a/core/sdk/src/extn/client/extn_client.rs
+++ b/core/sdk/src/extn/client/extn_client.rs
@@ -240,10 +240,20 @@ impl ExtnClient {
                                 });
                             } else {
                                 // could be main contract
-                                Self::handle_stream(message, self.request_processors.clone());
+                                if !Self::handle_stream(
+                                    message.clone(),
+                                    self.request_processors.clone(),
+                                ) {
+                                    self.handle_no_processor_error(message);
+                                }
                             }
                         } else {
-                            Self::handle_stream(message, self.request_processors.clone());
+                            if !Self::handle_stream(
+                                message.clone(),
+                                self.request_processors.clone(),
+                            ) {
+                                self.handle_no_processor_error(message);
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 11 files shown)


core/sdk/src/extn/client/extn_client.rs
```diff
@@ -240,10 +240,20 @@ impl ExtnClient {
                                 });
                             } else {
                                 // could be main contract
-                                Self::handle_stream(message, self.request_processors.clone());
+                                if !Self::handle_stream(
+                                    message.clone(),
```
Karthick-Somasundaresan, line 244: Is it possible to take ownership itself rather than cloning?

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -240,10 +240,20 @@ impl ExtnClient {
...
+                                    self.request_processors.clone(),
+                                ) {
+                                    self.handle_no_processor_error(message);
+                                }
                             }
                         } else {
-                            Self::handle_stream(message, self.request_processors.clone());
+                            if !Self::handle_stream(
+                                message.clone(),
```
Karthick-Somasundaresan, line 252: Same here as well. Is it possible to take ownership of the message rather than cloning it?

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -262,6 +272,21 @@ impl ExtnClient {
         debug!("Initialize Ended Abruptly");
     }
 
+    fn handle_no_processor_error(&self, message: ExtnMessage) {
+        let req_sender = if let Some(requestor_sender) =
+            self.get_extn_sender_with_extn_id(message.clone().requestor.to_string())
```
Karthick-Somasundaresan, line 277: Do we need to clone the message in order to get the requestor's string?

core/sdk/src/extn/client/extn_client.rs
satlead: It passes on the value down the processor in its own thread so it needs to get an instance. If we cant find a processor we send it to a different handler to inform the caller that there was no processor available to handle it.

core/sdk/src/extn/client/extn_client.rs
satlead: Same as above

Karthick-Somasundaresan (approved):

PR #177, opened 2023-07-25
New api support

Karthick-Somasundaresan:


satlead (approved):

core/main/src/state/firebolt-open-rpc.json
satlead, line 1: We cannot merge this change till we implement ripple rpc. Its missing lifecycle management API(s)

satlead (changes requested):

core/main/src/state/firebolt-open-rpc.json
satlead: Can we add ripple-rpc.json to the startup so its also there?

satlead (approved):
LGTM

PR #187, opened 2023-08-03
Improve setup process

adamdama:
## What

Improve the Ripple init process and update formatting in the README file.

## Why

Getting started with working on the project should be as simple as possible and so automating simple steps will ease friction for new users.

## How

A few of the steps in the getting started part of the README can be automated. This PR updates the ./ripple script's init functionality to set values in the manifest files that can be automatically determined. Backing up existing .ripple folder has also been automated.

Additionally, the init script now also tells the user what is happening so that they know it has worked.

### DIFF
--- a/ripple
+++ b/ripple
@@ -35,7 +35,10 @@ function get_default_extension() {
         Darwin)
             echo "dylib"
             ;;
-        *)
+        CYGWIN*|MINGW32*|MSYS*|MINGW*)
+            echo "dll"
+            ;;
+        Linux)
             echo "so"
             ;;
     esac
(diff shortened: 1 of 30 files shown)


ripple
```diff
@@ -30,13 +30,63 @@ function print_help() {
     echo "ripple run 10.0.0.1"
 }
 
+function get_default_extension() {
+    case "$(uname -s)" in
+        Darwin)
+            echo "dylib"
+            ;;
+        *)
+            echo "so"
```
satlead, line 39: Add Windows

adamdama: Done in https://github.com/rdkcentral/Ripple/pull/187/commits/7db403949b62cedf3c850505ac925b583682d3e6

satlead (approved):
Ty

Issue #192, opened 2023-08-04
Fix warnings raised by Clippy


adamdama:
The Rust Lang team provide a Rust code linting tool called [Clippy](https://github.com/rust-lang/rust-clippy). The tool can be used to check for common mistakes in a Rust code base and improve the quality of implemented code. The tool is an invaluable resource to Rust developers and enables collaborators to learn recommended implementation patterns as well as pickup things which are not easy to see in a review, such as redundant calls to `.clone()`. Clippy even offers a `--fix` argument to automatically fix some of the issues identified by the linter.

Running Clippy against the existing code base currently returns 391 warnings which are things that can be easily addressed. 

I am going to submit a PR that will add a local git pre-commit hook to run Clippy before new code is committed but that will mean that all of these warnings will be printed into the terminal for each commit made. Therefore I think we should address them before this hook is implemented to avoid a less than ideal experience for contributors.

The current warnings can be seen by running `clippy` in the root of the project:
```
cargo clippy
```

PR #194, opened 2023-08-04
Clippy fixes: auto and manual app_launcher

adamdama:
## What

Refactoring code that triggers clippy warnings into the recommended code. Automatic fixes and manual fixes for 1 file only.

## Why

To start clear the current warning list presented when running clippy against the code base.

https://github.com/rdkcentral/Ripple/issues/192

## How

By running `cargo clippy --fix`.
By manually resolving all warnings in core/launcher/src/manager/app_launcher.rs

### DIFF
--- a/core/main/src/firebolt/handlers/advertising_rpc.rs
+++ b/core/main/src/firebolt/handlers/advertising_rpc.rs
@@ -52,7 +52,7 @@ use super::{
     privacy_rpc::{self, PrivacyImpl},
 };
 
-const ADVERTISING_APP_BUNDLE_ID_SUFFIX: &'static str = "Comcast";
+const ADVERTISING_APP_BUNDLE_ID_SUFFIX: &str = "Comcast";
 
 #[derive(Serialize, Debug)]
 #[serde(rename_all = "camelCase")]
@@ -141,7 +141,7 @@ impl AdvertisingServer for AdvertisingImpl {
             .send_extn_request(AdvertisingRequest::ResetAdIdentifier(session))
             .await;
 
-        if let Err(_) = resp {
+        if resp.is_err() {
             error!("Error resetting ad identifier {:?}", resp);
             return Err(rpc_err("Could not reset ad identifier for the device"));
         }
@@ -162,12 +162,12 @@ impl AdvertisingServer for AdvertisingImpl {
         let payload = AdvertisingRequest::GetAdIdObject(AdIdRequestParams {
             privacy_data: privacy_rpc::get_allow_app_content_ad_targeting_settings(&self.state)
                 .await,
-            app_id: ctx.clone().app_id.to_string(),
+            app_id: ctx.clone().app_id,
             dist_session: session,
         });
         let resp = self.state.get_client().send_extn_request(payload).await;
 
-        if let Err(_) = resp {
+        if resp.is_err() {
             error!("Error getting ad init object: {:?}", resp);
             return Err(rpc_err("Could not get ad init object from the device"));
         }
@@ -232,7 +232,7 @@ impl AdvertisingServer for AdvertisingImpl {
 
         let resp = self.state.get_client().send_extn_request(payload).await;
 
-        if let Err(_) = resp {
+        if resp.is_err() {
             error!("Error getting ad init object: {:?}", resp);
             return Err(rpc_err("Could not get ad init object from the device"));
         }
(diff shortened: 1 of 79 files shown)


device/thunder_ripple_sdk/src/processors/thunder_persistent_store.rs
```diff
@@ -118,11 +118,11 @@ impl ThunderStorageRequestProcessor {
             .get_thunder_client()
             .call(DeviceCallRequest {
                 method: thunder_method,
-                params: params,
+                params,
             })
             .await;
         if response.message.get("success").is_none()
-            || response.message["success"].as_bool().unwrap() == false
+            || !response.message["success"].as_bool().unwrap()
```
Karthick-Somasundaresan, line 125: Since we are touching this part, can we replace unwrap with unwrap_or_default() which is safer.

device/thunder_ripple_sdk/src/processors/thunder_persistent_store.rs
```diff
@@ -92,11 +92,11 @@ impl ThunderStorageRequestProcessor {
...
```
Karthick-Somasundaresan, line 99: Since we are touching this part, can we replace unwrap with unwrap_or_default() which is safer.

core/main/src/firebolt/handlers/advertising_rpc.rs
```diff
@@ -162,12 +162,12 @@ impl AdvertisingServer for AdvertisingImpl {
         let payload = AdvertisingRequest::GetAdIdObject(AdIdRequestParams {
             privacy_data: privacy_rpc::get_allow_app_content_ad_targeting_settings(&self.state)
                 .await,
-            app_id: ctx.clone().app_id.to_string(),
+            app_id: ctx.clone().app_id,
```
Karthick-Somasundaresan, line 165: Rather than cloning the ctx to get app_id, I think we can do ctx.app_id.to_string() or ctx.app_id.to_owned()

core/main/src/firebolt/handlers/capabilities_rpc.rs
```diff
@@ -140,19 +140,17 @@ impl CapabilityServer for CapabilityImpl {
             .check_cap_role(&ctx.app_id, cap.clone())
         {
             return Ok(v);
-        } else {
-            if let Ok(_) =
-                PermissionHandler::fetch_and_store(self.state.clone(), ctx.clone().app_id).await
+        } else if let Ok(_) =
+            PermissionHandler::fetch_and_store(self.state.clone(), ctx.clone().app_id).await
```
Karthick-Somasundaresan, line 144: Here as well. We can get a string from app_id rather than cloning context and getting app_id from it.

Karthick-Somasundaresan: A better approach would be changing the function fetch_and_store to get string slice rather than an owned string.

core/main/src/firebolt/handlers/capabilities_rpc.rs
```diff
@@ -304,16 +302,16 @@ pub async fn is_permitted(
...
-            //successful fetch retry
-            if let Ok(v) = state
-                .cap_state
-                .permitted_state
-                .check_cap_role(&ctx.app_id, cap)
-            {
-                return Ok(v);
-            }
+    } else if let Ok(_) =
+        PermissionHandler::fetch_and_store(state.clone(), ctx.clone().app_id).await
```
Karthick-Somasundaresan, line 306: If we change the function signature we can avoid the second clone here as well. In fact, we can even make the function take reference for platform state and app id so that we can avoid both clones in this function.

Karthick-Somasundaresan (changes requested):
Great work in capturing these errors. Thank you. In addition to that I had given a few minor code review comments. Kindly could you please address them?

satlead:
Think the above changes are mostly done by the Tool and it doesn't seem to refactor(which is probably Ok). 
@adamdama is planning to integrate the Clippy tool into our local git hooks so every commit is validated.
My suggestion is to create separate issues for these and go forward with removing warnings and errors so the git hook becomes ready.

These new issues would also be good training for anyone to start off their contributions to Ripple.

Karthick-Somasundaresan:
> Think the above changes are mostly done by the Tool and it doesn't seem to refactor(which is probably Ok). @adamdama is planning to integrate the Clippy tool into our local git hooks so every commit is validated. My suggestion is to create separate issues for these and go forward with removing warnings and errors so the git hook becomes ready.
> 
> These new issues would also be good training for anyone to start off their contributions to Ripple.

Agreed.

PR #202, opened 2023-08-14
feat: Metric Upgrades

satlead:
## What
Metrics upgrades to support
1. Extensions to send metrics and telemetry information
2. Example extension for developers to create their own telemetry listener
3. Provide support for extensions to create their own Proprietary Metrics schema for consumption.
4. Implement metrics management Ripple RPC Api

## Why
To offer better metrics support for extensions and better encapsulation of Ripple state information.
Extensions should be agnostic to the internal state of the Main application but offer enough metadata for Metrics.
Proprietary extensions should also have the ability to create their own Metrics schema using RawMetric enumeration.

## How
1. Contracts for Operational Telemetry listeners and Metrics Forwarder
2. Operational Telemetry Support using Telemetry Builder for extensions to receive telemetry events
3. Sample extension for receiving the telemetry events and forwarding them to a WebSocket
4. Proprietary Metrics support using `serde_json::Value`, where Main will populate Ripple State information as it passes through.
5. Add RPC handler for MetricsManagement

### DIFF
--- a/core/main/src/bootstrap/extn/start_extn_channel_step.rs
+++ b/core/main/src/bootstrap/extn/start_extn_channel_step.rs
@@ -15,12 +15,14 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use std::time::Duration;
+
 use ripple_sdk::{
     api::status_update::ExtnStatus,
     async_trait::async_trait,
     framework::{bootstrap::Bootstep, RippleResponse},
     log::error,
-    tokio::sync::mpsc,
+    tokio::{sync::mpsc, time::timeout},
     utils::error::RippleError,
 };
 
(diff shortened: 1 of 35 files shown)


joseinweb (approved):

core/main/src/bootstrap/extn/start_extn_channel_step.rs
```diff
@@ -82,11 +84,20 @@ impl Bootstep<BootstrapState> for StartExtnChannelsStep {
                 .extn_state
                 .add_extn_status_listener(extn_id.clone(), tx)
             {
-                if let Some(v) = tr.recv().await {
-                    state.extn_state.clear_status_listener(extn_id);
-                    match v {
-                        ExtnStatus::Ready => continue,
-                        _ => return Err(RippleError::BootstrapError),
+                match timeout(Duration::from_millis(10000), tr.recv()).await {
```
Karthick-Somasundaresan, line 87: Is it okay to have such a long waiting period blocking the boot sequence?

satlead: The extensions can be implemented to

1. Connect to Thunder or underlying API, 
2. Setup for pub-sub connections 

Want to start with a big number and then get values across all devices to come up with a reasonable threshold. 
This is the worst-case scenario in most cases the extensions start less than 300 mSecs with all the loaded functionality in parallel.

The total bootstrap step loading all extensions with device and distributor connections takes less than 300 msecs.
```
2023-08-16-08:28:05.311[DEBUG][ripple_sdk::framework::bootstrap][gateway]->>>Starting Bootstep StartExtnChannelsStep<<<
2023-08-16-08:28:05.807[DEBUG][ripple_sdk::framework::bootstrap][gateway]----Successful Bootstep StartExtnChannelsStep---
```

Issue #199, opened 2023-08-09
Loop that never loops


adamdama:
During the process of addressing clippy warnings on the code base I found a for loop that never actually loops. I have attached screenshots of the code block and error. Code is [here](https://github.com/rdkcentral/Ripple/blob/f466ae6bc75dd3bc26a5a9a41dfc5e272b505e2f/core/main/src/service/user_grants.rs#L897)

![Screenshot 2023-08-09 at 09 15 30](https://github.com/rdkcentral/Ripple/assets/2086479/c7baa2ab-f9b8-4848-8f80-b19e98f65b11)
![Screenshot 2023-08-09 at 09 15 37](https://github.com/rdkcentral/Ripple/assets/2086479/c3d5211d-2295-427c-8653-5013ed938efb)

If you look at the branches in the inner loop all of them contain a `return` statement, therefore the loop will exit on the first iteration no matter what. I can of course refactor this to use `.iter().next()` and just handle the first case but I am concerned that this is not the intention of this code.

It would be great to get some guidance on how the `evaluate_options` function is supposed to work so that I can address the problem in the most appropriate way.

satlead:
Believe the if condition should be moved outside the loop as we already have the info we need to call that condition.

PR #203, opened 2023-08-15
fix: user grant option evaluation

adamdama:
## What

Fixed evaluation of user grant policy options. This addresses the issue raised [here](https://github.com/rdkcentral/Ripple/issues/199)

This is also the final warning/error detected by clippy which addresses the issue raised [here](https://github.com/rdkcentral/Ripple/issues/192)

## Why

The current implementation of the user grant options evaluation only looks at the first step in the first option of the grant policies set. This option has its first step executed even if the platform does not support it. The result of this execution is then used as the entire result of the options evaluation.

What should actually happen is the first option with all supported steps is executed and a grant/deny response produced from a combination of capability support and user actions.

## How

The evaluate options loop has been broken up into a multi step process where we first find the option which is supported by the platform. This option is then used to drive the execution of the grant steps, which prompts the user for the appropriate input. The outcome of this process is then used to generate an overall response as to whether or not the use of the capability is approved.

Unit test cases have been added to cover all of the combinations of cases to ensure that this process is followed correctly.

### DIFF
--- a/core/tdk/src/utils/test_utils.rs
+++ b/core/tdk/src/utils/test_utils.rs
@@ -15,6 +15,16 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use serde_json::json;
+
 pub trait Mockable {
     fn mock() -> Self;
+}
+
+pub fn cap_jsonrpc_payload_granted(cap: String) -> serde_json::Value {
+    json!({"id":1,"jsonrpc":"2.0","result":{"available":true,"capability":cap,"manage":{"granted":true,"permitted":true},"provide":{"granted":true,"permitted":true},"supported":true,"use":{"granted":true,"permitted":true}}})
+}
+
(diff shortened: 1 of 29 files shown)


core/tdk/src/utils/test_utils.rs
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use serde_json::json;
+
+pub trait Mockable {
+    fn mock() -> Self;
+}
+
+pub fn cap_jsonrpc_payload_granted(cap: String) -> serde_json::Value {
```
Karthick-Somasundaresan, line 24: In future we can expand this function to take vec of Role as well so that we can use this function to test for more test cases altering granted status for different roles. For now this is fine.

Karthick-Somasundaresan (approved):

PR #208, opened 2023-08-18
feat: Adjectives ADR + Implementation

satlead:
## What

Provide a better design for contracts

## Why
https://github.com/rdkcentral/Ripple/blob/adjectives/docs/adr/contract_adjectives.md

## How
https://github.com/rdkcentral/Ripple/blob/adjectives/docs/adr/contract_adjectives.md

### DIFF
--- a/core/sdk/src/api/device/device_peristence.rs
+++ b/core/sdk/src/api/device/device_peristence.rs
@@ -17,7 +17,7 @@
 
 use crate::{
     extn::extn_client_message::{ExtnPayload, ExtnPayloadProvider, ExtnRequest},
-    framework::ripple_contract::RippleContract,
+    framework::ripple_contract::RippleContract, api::storage_property::StorageAdjective,
 };
 use chrono::Utc;
 use serde::{Deserialize, Serialize};
@@ -93,6 +93,6 @@ impl ExtnPayloadProvider for DevicePersistenceRequest {
     }
 
     fn contract() -> RippleContract {
-        RippleContract::DevicePersistence
+        RippleContract::Storage(StorageAdjective::Local)
     }
 }
--- a/core/sdk/src/api/device/device_peristence.rs
+++ b/core/sdk/src/api/device/device_peristence.rs
@@ -16,8 +16,9 @@
 //
 
 use crate::{
+    api::storage_property::StorageAdjective,
     extn::extn_client_message::{ExtnPayload, ExtnPayloadProvider, ExtnRequest},
-    framework::ripple_contract::RippleContract, api::storage_property::StorageAdjective,
+    framework::ripple_contract::RippleContract,
 };
 use chrono::Utc;
 use serde::{Deserialize, Serialize};
--- a/core/sdk/src/extn/client/extn_client.rs
+++ b/core/sdk/src/extn/client/extn_client.rs
@@ -128,19 +128,20 @@ impl ExtnClient {
         };
         let mut atleast_one_contract_satisfied = false;
         let mut contracts_supported = Vec::new();
-        for contract in contracts {
-            if self.sender.check_contract_fulfillment(contract) {
-                atleast_one_contract_satisfied = true;
+        let contracts_supported = contracts
+            .iter()
+            .filter(|contract| self.sender.check_contract_fulfillment(contract))
+            .collect();
+            
+        contracts_supported.for_each(|contract| {
                 let processor_string: String = processor.contract().as_clear_string();
                 info!("adding request processor {}", processor_string);
                 add_stream_processor(
-                    processor_string.clone(),
+                    processor_string,
                     processor.sender(),
                     self.request_processors.clone(),
                 );
-                contracts_supported.push(processor_string);
-            }
-        }
+        });
         // Dont add and start a request processor if there is no contract fulfillment
         if atleast_one_contract_satisfied {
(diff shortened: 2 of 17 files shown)


core/sdk/src/extn/client/extn_client.rs
```diff
@@ -121,17 +121,33 @@ impl ExtnClient {
     ///
     /// Also starts the thread in the processor to accept incoming requests.
     pub fn add_request_processor(&mut self, mut processor: impl ExtnRequestProcessor) {
+        let contracts = if let Some(multiple_contracts) = processor.fulfills_mutiple() {
+            multiple_contracts
+        } else {
+            vec![processor.contract()]
+        };
+        let mut atleast_one_contract_satisfied = false;
```
adamdama, line 129: Do you need this variable? Is it not the same as `!contracts_supported.is_empty()`

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -121,17 +121,33 @@ impl ExtnClient {
...
+                let processor_string: String = processor.contract().as_clear_string();
+                info!("adding request processor {}", processor_string);
+                add_stream_processor(
+                    processor_string.clone(),
+                    processor.sender(),
+                    self.request_processors.clone(),
+                );
+                contracts_supported.push(processor_string);
+            }
+        }
```
adamdama, line 143: Rather than use a mutable vector to collect the contracts we could filter the contracts to find the suitable ones and then add them all as stream_processers.

```suggestion
        let contracts_supported = contracts
            .iter()
            .filter(|contract| self.sender.check_contract_fulfillment(contract))
            .collect();
            
        contracts_supported.for_each(|contract| {
                let processor_string: String = processor.contract().as_clear_string();
                info!("adding request processor {}", processor_string);
                add_stream_processor(
                    processor_string,
                    processor.sender(),
                    self.request_processors.clone(),
                );
        });
```

I am not sure if this suits the code style we are working towards but I find reduced nesting a little easier to read

core/sdk/src/framework/ripple_contract.rs
```diff
@@ -121,7 +133,41 @@ impl RippleContract {
...
+        }
+
+        None
+    }
+
+    pub fn from_adjective_string(contract: &str, adjective: &str) -> Option<Self> {
+        let adjective = format!("\"{}\"", adjective);
+        match contract {
+            "storage" => {
+                if let Ok(v) = serde_json::from_str::<StorageAdjective>(&adjective) {
```
adamdama, line 159: cxIf this does not deserialise and we get an Err should we at least log it? At that point we will return None and we may never know there is a problem.

core/sdk/src/framework/ripple_contract.rs
```diff
@@ -178,4 +226,25 @@ mod tests {
...
+
+        // other way around
+        let manifest_entry = String::from("local.storage");
+        let result = RippleContract::try_from(manifest_entry);
+        assert!(result.is_ok());
+        assert!(matches!(
+            result,
+            Ok(RippleContract::Storage(StorageAdjective::Local))
+        ));
+    }
```
adamdama, line 249: Can we just make two test cases here?

docs/adr/contract_adjectives.md
```diff
@@ -0,0 +1,273 @@
...
+#### Storage
+1. I want to store a **privacy local** value.
+2. I want to store a **privacy cloud** value.
+3. I want to store a **usergrant local** value.
+4. I want to store a **usergrant cloud** value.
+5. I want to store a **local device** value.
+
+#### Metrics
+1. I want to forward **behavior** metrics.
+2. I want to forward **behavior** metrics.
```
adamdama, line 88: This point is the same as the first one.

adamdama (commented):
I think this is a very elegant solution to the problem. The ADR makes and offers invaluable context to the changes. Awesome stuff.

I have left a few comments about code refactors, but I see no real reason why this couldn't be merged.

core/sdk/src/api/device/device_peristence.rs
```diff
@@ -93,6 +94,6 @@ impl ExtnPayloadProvider for DevicePersistenceRequest {
     }
 
     fn contract() -> RippleContract {
-        RippleContract::DevicePersistence
+        RippleContract::Storage(StorageAdjective::Local)
```
pahearn73, line 97: General question: Is there currently anything preventing a contract from being associated with multiple "adjectives" if needed in the future?

adamdama (approved):

core/sdk/src/extn/client/extn_client.rs
satlead: Actually they depend on whether this contract is enabled on the extension manifest. This uses `check_contract_fulfillment()` method to see if a given extension has the entry in manifest. If not it will be ignored and not setup.

satlead: Added your patch

core/sdk/src/api/device/device_peristence.rs
satlead: Yes currently contracts dont have adjectives and they often have one to one mapping with a request and a processor so its not very scalable. Adjectives solves all those dependency issues

adamdama (approved):

PR #209, opened 2023-08-21
178 launcher extension needs to pass on session

kvfasil:
What
Currently, Launcher Extension doesn't use the sessionId received back from Main and pass on to the app being launched.

Why
These changes were not implemented in lieu with just being used as a Certification platform. With Accelerator UI effort Ripple will be using Launcher for more than just certification so we need to launch apps with session id to the secure port.

### DIFF
--- a/core/launcher/Cargo.toml
+++ b/core/launcher/Cargo.toml
@@ -30,3 +30,4 @@ ripple_sdk = { path = "../../core/sdk" }
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 url = "2.2.2"
+urlencoding = "2.1.0"
--- a/core/launcher/src/manager/app_launcher.rs
+++ b/core/launcher/src/manager/app_launcher.rs
@@ -43,7 +43,7 @@ use ripple_sdk::{
         manifest::{app_library::AppLibrary, apps::AppManifest, device_manifest::RetentionPolicy},
     },
     extn::extn_client_message::ExtnResponse,
-    framework::RippleResponse,
+    framework::{ripple_contract::RippleContract, RippleResponse},
     log::{debug, error, info, warn},
     tokio::{
         self,
@@ -67,6 +67,7 @@ use super::{
 use ripple_sdk::futures::future::{BoxFuture, FutureExt};
 
 pub const NAVIGATION_INTENT_PROVIDER_REQUEST: &str = "providerRequest";
+pub const BRIDGEPROTOCOL: &str = "BridgeProtocol";
 #[derive(Debug, Clone)]
 struct App {
     #[allow(dead_code)]
@@ -513,8 +514,12 @@ impl AppLauncher {
         }
     }
 
-    fn get_transport(url: String) -> AppRuntimeTransport {
-        if !url.as_str().contains("__firebolt_endpoint") {
+    fn get_transport(state: &LauncherState, url: String) -> AppRuntimeTransport {
+        if !url.as_str().contains("__firebolt_endpoint")
+            && state
+                .extn_client
+                .check_contract_permitted(RippleContract::BridgeProtocol)
+        {
             AppRuntimeTransport::Bridge
         } else {
             AppRuntimeTransport::Websocket
@@ -536,7 +541,7 @@ impl AppLauncher {
             },
             runtime: Some(AppRuntime {
                 id: Some(callsign),
-                transport: Self::get_transport(manifest.start_page),
+                transport: Self::get_transport(&state, manifest.start_page),
             }),
             launch: AppLaunchInfo {
                 intent: Some(intent),
--- a/core/launcher/src/manager/app_launcher.rs
+++ b/core/launcher/src/manager/app_launcher.rs
@@ -53,6 +53,7 @@ use ripple_sdk::{
 };
 use serde_json::Value;
 use url::Url;
+use urlencoding::encode;
 
 use crate::{
     launcher_state::LauncherState,
@@ -514,7 +515,7 @@ impl AppLauncher {
         }
     }
 
-    fn get_manifest_url(manifest: AppManifest) -> String {
+    fn get_modified_url(manifest: AppManifest, sessionid: &str) -> String {
         let mut url = manifest.start_page.to_string();
         if let Ok(mut modified_url) = Url::parse(&url) {
             let mut query_params: Vec<(String, String)> = modified_url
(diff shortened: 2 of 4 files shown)


core/launcher/src/manager/app_launcher.rs
```diff
@@ -513,8 +514,34 @@ impl AppLauncher {
...
+        if !(url.clone().as_str().contains("__firebolt_endpoint")) {
+            if let Ok(mut parsed_url) = Url::parse(url.clone().as_str()) {
+                let firebolt_endpoint = String::from("127.0.0.0");
+                let appid = manifest.name.to_string();
+
+                let modified_query: String = format!(
+                    "__firebolt_endpoint=ws%3A%2F%2F{}%3A3473%3FappId%3D{}",
+                    firebolt_endpoint, appid
+                );
+                parsed_url.set_query(Some(&modified_query));
```
satlead, line 528: This will override the query for all the url(s). 
For eg if an app url has
```
https://cool.app?neededParam=foo
```

It will change to 
```
https://cool.app?__firebolt_endpoint=ws%3A%2F%2F127.0.0.1%3A3473%3FappId%3D
```
Apps would fail if some mandatory params went missing.
My suggestion is to append to an existing url param before set_query

core/launcher/src/manager/app_launcher.rs
```diff
@@ -545,16 +572,39 @@ impl AppLauncher {
...
+                    return Err(AppError::NotFound);
+                }
+            },
+            Err(_) => {
+                return Err(AppError::NotSupported);
+            }
         }
-        Ok(())
+
+        info!("modified url with sessionid : {:?}", modified_url);
```
satlead, line 606: Probably a Debug or Trace not an Info

satlead (changes requested):
Can we add a Unit test for the get_transport method for all 3 cases

1. Already has __firebolt_endpoint param
2. Doesnt have __firebolt_endpoint param or any other params
3. Doesnt have __firebolt_endpoint param but has other params

core/launcher/src/manager/app_launcher.rs
kvfasil: changes added in latest commit.

core/launcher/src/manager/app_launcher.rs
kvfasil: debug added.

core/launcher/src/manager/app_launcher.rs
```diff
@@ -513,8 +514,50 @@ impl AppLauncher {
         }
     }
 
-    fn get_transport(url: String) -> AppRuntimeTransport {
-        if !url.as_str().contains("__firebolt_endpoint") {
+    fn get_manifest_url(manifest: AppManifest) -> String {
```
satlead, line 517: Can we add some unit tests

satlead (dismissed):

core/launcher/Cargo.toml
```diff
@@ -30,3 +30,4 @@ ripple_sdk = { path = "../../core/sdk" }
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 url = "2.2.2"
+urlencoding = "2.1.0"
```
satlead, line 33: @joseinweb we need to add this in the dependency

core/launcher/src/manager/app_launcher.rs
```diff
@@ -513,8 +515,47 @@ impl AppLauncher {
         }
     }
 
-    fn get_transport(url: String) -> AppRuntimeTransport {
-        if !url.as_str().contains("__firebolt_endpoint") {
+    fn get_modified_url(manifest: AppManifest, sessionid: &str) -> String {
```
satlead, line 518: Better testable

core/launcher/src/manager/app_launcher.rs
```diff
@@ -840,3 +902,53 @@ impl AppLauncher {
...
+        assert_eq!(modified_url, "http://example.com/?param1=value1&__firebolt_endpoint=ws%3A%2F%2F127.0.0.0%3A3473%26appId%3Dtest%3Fsession%3Dmy_session_id#menu");
+    }
+
+    #[test]
+    fn test_get_modified_url_with_firebolt_endpoint() {
+        let mut app_lib = AppManifest::default();
+        app_lib.start_page = String::from("http://example.com/?param1=value1&__firebolt_endpoint=ws%253A%252F%252F127.0.0.0%253A3473%253FappId%253Dmy_app_id");
+        let session_id = "my_session_id";
+        let modified_url = AppLauncher::get_modified_url(app_lib, session_id);
+        assert_eq!(modified_url, modified_url);
```
satlead, line 943: Fix this

core/launcher/src/manager/app_launcher.rs
kvfasil: done

core/launcher/src/manager/app_launcher.rs
```diff
@@ -513,8 +515,43 @@ impl AppLauncher {
...
+                    .join("&"),
+            ));
+            modified_url.as_str().to_string()
+        } else {
+            debug!("Invalid URL");
+            "Invalid URL".to_string()
+        }
+    }
+
+    fn get_transport(contract_permited: bool, url: String) -> AppRuntimeTransport {
```
Karthick-Somasundaresan, line 553: url could have been taken as a reference rather than as an owned one. This avoids cloning in the calling place.

kvfasil: done

core/launcher/src/manager/app_launcher.rs
```diff
@@ -513,8 +515,43 @@ impl AppLauncher {
         }
     }
 
-    fn get_transport(url: String) -> AppRuntimeTransport {
-        if !url.as_str().contains("__firebolt_endpoint") {
+    fn get_modified_url(manifest: AppManifest, sessionid: &str) -> String {
+        let url = manifest.start_page.to_string();
```
Karthick-Somasundaresan, line 519: I think this is not required, anyway, we have to pass a reference to Url::parse. so we can call it as Url::parse(&manifest.start_page)

core/launcher/src/manager/app_launcher.rs
```diff
@@ -513,8 +515,43 @@ impl AppLauncher {
         }
     }
 
-    fn get_transport(url: String) -> AppRuntimeTransport {
-        if !url.as_str().contains("__firebolt_endpoint") {
+    fn get_modified_url(manifest: AppManifest, sessionid: &str) -> String {
```
Karthick-Somasundaresan, line 518: I don't think we need an owned manifest. Cloning is a costly operation. We have to avoid it as much as possible. Here in this function, we are only reading the values from manifest. So I think we can take that as a reference.

Karthick-Somasundaresan (changes requested):

core/launcher/src/manager/app_launcher.rs
kvfasil: changes added.

core/launcher/src/manager/app_launcher.rs
kvfasil: done

Karthick-Somasundaresan (approved):

PR #214, opened 2023-08-21
Clippy integration

maggie98choy:


core/main/src/service/user_grants.rs
adamdama, line 1: I think the changes in this loop are not correct. There is an issue with the current user_grants evaluate_options implementation. Changing this to just process the first one makes the code operate the same as it does today but it hides the problem.

maggie98choy: We would probably need to revise the logic within the for..loop block.  Looking into the for-loop, we return from the loop in every condition.

Vinodsathyaseelan: @Karthick-Somasundaresan : Is this specific to Ripple 2.0 ?

adamdama: Yes the logic is currently incorrect. @Karthick-Somasundaresan and I have been addressing it here:
https://github.com/rdkcentral/Ripple/pull/203
https://github.com/rdkcentral/Ripple/issues/199

This should be merged today and then we can get this branch updated and finally have Clippy integration 🎉 ❤️

run-clippy.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -v
+echo "start to run cargo clippy"
+cargo clippy --fix --allow-staged >> output.txt
```
adamdama, line 5: Why don't we just print to stdout? Personally, I'd rather see issues in the console than look in a text file.

If we are to go with a test file then can we give it a more specific name and add it to .gitignore to prevent it from being committed by accident.

adamdama:
I have merged this PR into the other Clippy one so that it is all in one place: https://github.com/rdkcentral/Ripple/pull/216

PR #216, opened 2023-08-23
Clippy hooks

adamdama:


core/sdk/src/api/config.rs
```diff
@@ -124,3 +124,5 @@ impl ExtnPayloadProvider for LauncherConfig {
         RippleContract::Config
     }
 }
+
+pub fn something() {}
```
satlead, line 128: Do we need this method

adamdama: No we do not. It was a test for the pre-commit hook. 🤦 

Sorry, I didn't self-review this properly.

satlead (approved):

PR #226, opened 2023-09-07
Commit privacy settings to local on success from ESS

maggie98choy:
## What

Currently there is a long delay when a user selects a toggle in the privacy settings app as Ripple waits for the linchpin notification, like 4 seconds. The user is waiting a this time to continue.

## Why

To try and get this delay down, Ripple should not need to wait for linchpin in this case.

## How

if the call to ESS results in 200 status code then it is successful and Ripple should update its local cache. Change it so on success for the set to ESS:

- Persist to local store/cache
- Send out the onChanged firebolt notification for that setting (if it changed)

## Test
- Verify that privacy settings is updated to ESS
- Verify that privacy setting is updated locally
- Verify subscriber receives privacy setting changes notification 
- Verify linchpin notification is received
- Check no local update, no notification is sent after receive linchpin notification.

## Checklist

- [ X] I have self-reviewed this PR

### DIFF
--- a/core/main/src/firebolt/handlers/privacy_rpc.rs
+++ b/core/main/src/firebolt/handlers/privacy_rpc.rs
@@ -541,20 +541,22 @@ impl PrivacyImpl {
             }
             PrivacySettingsStorageType::Cloud | PrivacySettingsStorageType::Sync => {
                 if let Some(dist_session) = platform_state.session_state.get_account_session() {
-                    let request = PrivacyCloudRequest::SetProperty(SetPropertyParams {
-                        setting: property.as_privacy_setting().unwrap(),
-                        value,
-                        dist_session,
-                    });
-                    let result = platform_state.get_client().send_extn_request(request).await;
-                    if PrivacySettingsStorageType::Sync == privacy_settings_storage_type
-                        && result.is_ok()
-                    {
-                        let _ =
-                            StorageManager::set_bool(platform_state, property, value, None).await;
-                    }
-                    if result.is_ok() {
-                        return Ok(());
+                    if let Some(privacy_setting) = property.as_privacy_setting() {
+                        let request = PrivacyCloudRequest::SetProperty(SetPropertyParams {
+                            setting: privacy_setting,
+                            value,
+                            dist_session,
+                        });
+                        let result = platform_state.get_client().send_extn_request(request).await;
+                        if PrivacySettingsStorageType::Sync == privacy_settings_storage_type
+                            && result.is_ok()
+                        {
+                            let _ = StorageManager::set_bool(platform_state, property, value, None)
+                                .await;
+                        }
+                        if result.is_ok() {
+                            return Ok(());
+                        }
                     }
                 }
                 Err(jsonrpsee::core::Error::Custom(String::from(&format!(
--- a/core/main/src/firebolt/handlers/privacy_rpc.rs
+++ b/core/main/src/firebolt/handlers/privacy_rpc.rs
@@ -521,18 +521,22 @@ impl PrivacyImpl {
             PrivacySettingsStorageType::Local => {
                 let payload = PrivacySettingsStoreRequest::SetPrivacySettings(property, value);
                 let response = platform_state.get_client().send_extn_request(payload).await;
-                if let Ok(extn_msg) =  response {
+                if let Ok(extn_msg) = response {
                     match extn_msg.payload {
-                        ExtnPayload::Response(res) => {
-                            match res {
-                                ExtnResponse::None(_) => RpcResult::Ok(()),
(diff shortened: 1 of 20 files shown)


core/main/src/firebolt/handlers/privacy_rpc.rs
```diff
@@ -549,9 +553,22 @@ impl PrivacyImpl {
                     "PrivacySettingsStorageType::Cloud: Not Available",
                 )))
             }
-            // PrivacySettingsStorageType::Sync => Err(jsonrpsee::core::Error::Custom(String::from(
-            //     "PrivacySettingsStorageType::Sync: Unimplemented",
-            // ))),
+            PrivacySettingsStorageType::Sync => {
+                let dist_session = platform_state.session_state.get_account_session().unwrap();
```
Vinodsathyaseelan, line 557: Is it safe to unwrap() ?  Has the dist_session been validated ? 
In Ripple 1.0, we have a function named is_valid() which  checks if the distributor session has all the fields populated with valid values.

core/main/src/firebolt/handlers/privacy_rpc.rs
```diff
@@ -549,9 +553,22 @@ impl PrivacyImpl {
...
+                let request = PrivacyCloudRequest::SetProperty(SetPropertyParams {
+                    setting: property.as_privacy_setting().unwrap(),
```
Vinodsathyaseelan, line 559: unsafe unwrap() ?

core/main/src/firebolt/handlers/privacy_rpc.rs
Karthick-Somasundaresan, line 557: This unwrap is not guaranteed to be safe always. Can we please check if it is none?

core/main/src/firebolt/handlers/privacy_rpc.rs
Karthick-Somasundaresan, line 559: Same here as well.

core/main/src/firebolt/handlers/privacy_rpc.rs
```diff
@@ -549,9 +553,22 @@ impl PrivacyImpl {
...
+                    value,
+                    dist_session,
+                });
+                let result = platform_state.get_client().send_extn_request(request).await;
```
Karthick-Somasundaresan, line 563: If I understand things correctly, the cloud updating part is the same both in sync and cloud. Only we need an additional step in case of sync. So can we refactor similar to the below snippet, such that we don't have to write again the cloud updation part? This way we have only one place to update in case of changes in the way we store.

```
match storage_type{
    PrivacySettingsStorageType::Cloud | PrivacySettingsStorageType::Sync => {
        let mut result = cloud_updation_portion();
        if storage_type == PrivacySettingsStorageType::Sync && result.is_ok(){
            result = local_updation_portion();
        }
        result.map_err(|_|{
            Err(jsonrpsee::core::Error::Custom(String::from(
                    "PrivacySettingsStorageType::Sync: Not Available",
            )))
        })
     }
}
```

Karthick-Somasundaresan (changes requested):

core/main/src/firebolt/handlers/privacy_rpc.rs
maggie98choy: Updated the pr to check if account session and privacy setting is Some then go on create a request param base on these structs.
In Ripple2.0, all the fields in account session are required so we just need to handle if account session is None.

core/main/src/firebolt/handlers/privacy_rpc.rs
maggie98choy: yeah, this is a cleaner way to add additional step in the case of sync.

Karthick-Somasundaresan (approved):

core/main/src/firebolt/handlers/privacy_rpc.rs
```diff
@@ -521,37 +521,49 @@ impl PrivacyImpl {
...
                     match extn_msg.payload {
-                        ExtnPayload::Response(res) => {
-                            match res {
-                                ExtnResponse::None(_) => RpcResult::Ok(()),
-                                _ => RpcResult::Err(jsonrpsee::core::Error::Custom("Unable to fetch".to_owned())),
-                            }
-                        }
-                        _ => RpcResult::Err(jsonrpsee::core::Error::Custom("Unexpected response received from Extn".to_owned()))
+                        ExtnPayload::Response(res) => match res {
+                            ExtnResponse::None(_) => RpcResult::Ok(()),
```
Vinodsathyaseelan, line 527: I assume that this is the only valid response from Extn for this request.

Vinodsathyaseelan (approved):

PR #230, opened 2023-09-12
Ripple context

Karthick-Somasundaresan:
## What

This PR will add Ripple context (device activation status and internet connection status) to each of the extension clients
## Why

So that the extension client will not have to make calls to the main extension.
## How

main will listen for the thunder events for the distribution session and internet changes. Whenever there is a change, it will update its local ripple context and send the change over the crossbeam channel to other extn clients to update their 
## Test

It was tested by changing the value for distributor token and remove the internet connection to the router to which STB is connected. which results in the thunder events and there will logs about the update in the ripple context in all extensions that were loaded.
## Checklist


core/main/src/bootstrap/extn/load_session_step.rs
```diff
@@ -32,6 +36,35 @@ impl Bootstep<BootstrapState> for LoadDistributorValuesStep {
...
+        // Start a thread which will process the session token update event.
+        // Upon receving the event, if the ripple context is not set as active, set it to active.
+        //
+        //
+        let _resp = s
+            .platform_state
+            .get_client()
+            .send_extn_request(DeviceEventRequest {
+                event: DeviceEvent::DistributorSessionTokenChanged,
+                id: "internal".to_owned(),
```
satlead, line 51: do we need this internal? By default all extensions talk back to Main

core/main/src/bootstrap/setup_extn_client_step.rs
```diff
@@ -54,6 +55,9 @@ impl Bootstep<BootstrapState> for SetupExtnClientStep {
         client.add_request_processor(KeyboardProcessor::new(state.platform_state.clone()));
         client.add_event_processor(ExtnStatusProcessor::new(state.clone().extn_state));
         client.add_event_processor(AppEventsProcessor::new(state.platform_state.clone()));
+        // client.add_event_processor(DistTokenChangeProcessor::new(
```
satlead, line 58: Can we delete it?

core/main/src/processor/mod.rs
```diff
@@ -19,10 +19,12 @@ pub mod account_link_processor;
 pub mod app_events_processor;
 pub mod authorized_info_processor;
 pub mod config_processor;
+// pub mod dist_token_change_processor;
```
satlead, line 22: Same here

core/main/src/processor/network_event_processor.rs
```diff
@@ -0,0 +1,80 @@
...
+#[async_trait]
+impl ExtnEventProcessor for NetworkEventProcessor {
+    async fn process_event(
+        _state: Self::STATE,
+        _msg: ExtnMessage,
+        extracted_message: Self::VALUE,
+    ) -> Option<bool> {
+        match extracted_message.clone() {
+            NetworkState::Connected => {
+                println!("Network is connected");
```
satlead, line 72: Change logging. shouldn't this be used instead of just println?

core/main/src/state/platform_state.rs
```diff
@@ -63,7 +63,7 @@ use super::{
 pub struct PlatformState {
     extn_manifest: ExtnManifest,
     device_manifest: DeviceManifest,
-    ripple_client: RippleClient,
+    pub ripple_client: RippleClient,
```
satlead, line 66: there is a method which provides the ripple_client using `get_client`

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -227,7 +252,45 @@ impl ExtnClient {
...
+                            .iter()
+                            .map(|(_, v)| v)
+                            .cloned()
+                            .collect();
+                        let is_main = self.sender.get_cap().is_main();
+
+                        if Self::is_ripple_context(&message) {
+                            let mut ripple_context = self.ripple_context.write().unwrap();
+                            if let ExtnPayload::Event(ExtnEvent::DistributorTokenChange(
+                                _dist_token,
```
satlead, line 270: Can we update this `_dist_token` token in Account session?

satlead (changes requested):
Some minor comments overall the flow looks good

core/main/src/bootstrap/extn/load_session_step.rs
Karthick-Somasundaresan: Yes, this request is not coming from any app, it is initiated from Ripple itself for internal purposes. Hence chosen that name

satlead: Hmm still trying to understand the need for that id given its not unique. What happens if we remove that field wouldn't SDK understand its internal feel it should be an Option and only added if the id is going to be unique

core/sdk/src/extn/client/extn_client.rs
satlead: To add on we expect Session Token to be available on updation. Then we can certainly store it in the Context instead of storing just the `DistributorTokenAvailable` enumeration

core/sdk/src/api/session.rs
```diff
@@ -58,6 +60,7 @@ pub enum AccountSessionRequest {
     Get,
     Provision(ProvisionRequest),
     SetAccessToken(AccountSessionTokenRequest),
+    GetAccessToken,
```
satlead, line 63: `Get` already gets AccessToken do we need another entry in the enumeration?

Karthick-Somasundaresan: But that does not have an expiry information. Only with that, we can invalidate the token in the ripple context. Since we don't have an event originating from thunder on token expiry.

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
```diff
@@ -68,9 +68,13 @@ impl ThunderPoolStep {
         info!("Thunder client connected successfully");
         let _ = state.extn_client.event(ExtnStatus::Ready);
         let extn_client = state.extn_client.clone();
-        Ok(ThunderBootstrapStateWithClient {
+        let thunder_boot_strap_state_with_client = ThunderBootstrapStateWithClient {
             prev: state,
             state: ThunderState::new(extn_client, client),
-        })
+        };
+        thunder_boot_strap_state_with_client
```
satlead, line 75: Ref

core/main/src/bootstrap/extn/load_session_step.rs
satlead: Lets fix this by moving App id into fireboltAppEvent field

satlead:
Closing this given we have https://github.com/rdkcentral/Ripple/pull/258 merged

PR #232, opened 2023-09-15
feat: Do not show user grant if calling app is not in active state

maggie98choy:
## What

Do not show user grant challenge if calling app is not in active state

## Why

In order to avoid the user being prompted by apps that are not even seen on screen.

## How

Ripple should just deny any API calls that require a user grant when the app is in that state.
Return error code = CAPABILITY_NOT_PERMITTED (-40300) with an error message:
"Capability cannot be used when app is in {state} state due to requiring a user grant"

Nothing should be stored in grant state for this denial.

If the capability requires a user grant but it is already in the grant state, then use that grant state.

## Test

We need to check two separate paths for this story:

1. Caller from method invocation
2. Caller from lifecyclemanagement.session() during"evaluateAt": ["activeSession"] path

## Checklist

- [X] I have self-reviewed this PR


core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -361,13 +361,15 @@ impl DelegatedLauncherHandler {
...
             Some(perms_with_grants) => {
                 // Grants required, spawn a thread to handle the response from grants
-
                 let cloned_ps = self.platform_state.clone();
                 let cloned_app_id = app_id.to_owned();
                 tokio::spawn(async move {
                     let resolved_result = GrantState::check_with_roles(
                         &cloned_ps,
-                        &CallerSession { session_id: None },
+                        &CallerSession {
```
Vinodsathyaseelan, line 369: Is it better to use CallerSession::default() here ?

core/main/src/service/extn/ripple_client.rs
```diff
@@ -96,6 +103,28 @@ impl RippleClient {
...
+    pub async fn get_app_state(&self, app_id: String) -> Result<String, RippleError> {
+        let (app_resp_tx, app_resp_rx) = oneshot::channel::<Result<AppManagerResponse, AppError>>();
+        let app_request = AppRequest::new(AppMethod::State(app_id.to_string()), app_resp_tx);
+        if let Err(e) = self.send_app_request(app_request) {
+            error!("Send error for get_state {:?}", e);
+            return Err(RippleError::SendFailure);
+        }
+        let resp: Result<Result<AppManagerResponse, AppError>, jsonrpsee::core::Error> =
+            rpc_await_oneshot(app_resp_rx).await;
+        // if let Ok(result) = resp {
```
Vinodsathyaseelan, line 115: You may remove the commented-out code.

core/main/src/service/user_grants.rs
```diff
@@ -79,14 +80,14 @@ pub struct GrantState {
 impl GrantState {
     pub fn new(manifest: DeviceManifest) -> GrantState {
         let saved_dir = manifest.clone().configuration.saved_dir;
-        let device_grant_path = format!("{}device_grants", saved_dir);
+        let device_grant_path = format!("{}/device_grants", saved_dir);
```
Vinodsathyaseelan, line 83: Not backward compatible with Ripple 1.0

core/main/src/service/user_grants.rs
```diff
@@ -79,14 +80,14 @@ pub struct GrantState {
...
         let dev_grant_store = if let Ok(v) = FileStore::load(device_grant_path.clone()) {
             v
         } else {
             FileStore::new(device_grant_path, HashSet::new())
         };
 
-        let app_grant_path = format!("{}app_grants", saved_dir);
+        let app_grant_path = format!("{}/app_grants", saved_dir);
         let app_grant_store = if let Ok(v) = FileStore::load(app_grant_path.clone()) {
```
Vinodsathyaseelan, line 91: Not backward compatible with Ripple 1.0

Vinodsathyaseelan: Not related to this ticket, but where are we updating the user_grants to persistent storage via FileStore:sync() method ?

core/main/src/service/user_grants.rs
```diff
@@ -1056,9 +1060,48 @@ impl GrantStepExecutor {
...
+                        debug!("App is not in foreground state");
+                        return Err(DenyReasonWithCap {
+                            reason: DenyReason::AppNotInActiveState,
+                            caps: vec![permission.cap.clone()],
+                        });
+                    }
+                }
+                Err(_) => {
+                    error!("Unable to get app state");
+                    return Err(DenyReasonWithCap {
```
Vinodsathyaseelan, line 1091: What is reason for returning `DenyReason::AppNotInActiveState` for this case ?

core/main/src/service/user_grants.rs
```diff
@@ -1357,7 +1401,7 @@ mod tests {
                     }],
                 },
             ]);
-            let caller_session: CallerSession = ctx.clone().into();
+            let caller_session: CallerSession = CallerSession::default();
```
Vinodsathyaseelan, line 1404: Passing None for app_id and session Id as caller_session for evaluate_options() ?  Why do we need to pass these default values?  
We have eliminated this dependency in Ripple 1.0.  Create an R2 Sync ticket to track those changes.

maggie98choy: This changes is in unit test. Passing None for app_id and session_id is to skip checking on app state because get_app_state is not mock to return foreground state therefore a few unit tests are failing.

Vinodsathyaseelan (dismissed):

core/main/src/service/user_grants.rs
Vinodsathyaseelan: The backward compatibility issue will be handled in a separate user story.

core/main/src/service/user_grants.rs
satlead: Created a new ticket for migrating the changes based on discussion with @Vinodsathyaseelan  https://github.com/rdkcentral/Ripple/issues/243

adamdama: If we use Path joining instead of string formatting, would this be compatible in both 1.0 and 2.0? I think generally when we are writing code to make fs paths we should use Path / PathBuf and not fromat!

satlead: Agreed `FileStore` uses Path during the loading stage but we can also create the path using the same class

Swati-Chand (dismissed):

satlead (approved):

Swati-Chand (approved):

PR #238, opened 2023-09-20
feat: mock device extension

adamdama:
## What

A ripple extension that supports setting up a WebSocket server to stand in for a device.
Thunder connection retries during boot

## Why

To support unit, component and (contract) provider verification testing in a containerised or desktop environment without access to a hardware device.

Additionally to support development of certain features which do not rely on a real device to function but require certain responses from the lower layers of the software stack.

## How

The premise of the mock device is simple. A WebSocket server with a registry of known requests and the responses we expect for each of those requests. When a request is made to the WebSocket server the incoming request is compared to the registry and if found the corresponding responses are sent back. If the request is unrecognised then nothing is sent back.

The extension is presented in two parts: a channel and a JSON-RPC extension. The channel holds the running WebSocket server that acts as the device and the JSON-RPC extension offers a control plane for the data within the server.

During extension loading a WebSocket server is started listening on the same port that the Thunder extension expects Thunder to be present on. In order to support this, connection retries have added to the Thunder SDK so that if the WebSocket server is not available immediately additional attempts to connect will be made.

The extension loads the initial state from a JSON file found in at a predetermined location. This is essential to get Ripple to boot up without crashing as certain requests and made to a device immediately after establishing the connection.

Once Ripple is running there are JSON-RPC methods which can be used to:
- add or overwrite requests in the mock data registry
- remove requests from the registry
- ~~emit an event~~ not yet implmented

## Test

Set up a manifest as per the included docs.
See docs for usage and try some of the patterns.

## Checklist


core/sdk/src/api/mod.rs
```diff
@@ -22,6 +22,7 @@ pub mod caps;
 pub mod config;
 pub mod device;
 pub mod manifest;
+pub mod mock_websocket_server;
```
satlead, line 25: There is an existing feature flag for `test`, feel we should add this as part of the feature flag?
Can this be added into `core/tdk` and be available for all test scripts?

core/sdk/src/framework/ripple_contract.rs
```diff
@@ -113,6 +113,8 @@ pub enum RippleContract {
     Metrics,
     /// Contract for Extensions to recieve Telemetry events from Main
     OperationalMetricListener,
+    /// Contract for Extensions to stand in for a WebSocket server based service provider
+    MockWebsocketServer,
```
satlead, line 117: Whats your thought on making `mock` as a common contract with MockTestAdjective?

you can then do
`ws.mock`
`http.mock`
`grpc.mock`
`com_rpc.mock` -

device/mock_device/src/mock_device_ffi.rs
```diff
@@ -0,0 +1,152 @@
...
+};
+
+const EXTN_NAME: &str = "mock_device";
+
+fn init_library() -> CExtnMetadata {
+    let _ = init_logger(EXTN_NAME.into());
+
+    let mock_device_channel = ExtnSymbolMetadata::get(
+        ExtnId::new_channel(ExtnClassId::Device, EXTN_NAME.into()),
+        ContractFulfiller::new(vec![RippleContract::MockWebsocketServer]),
```
satlead, line 53: Can a given extension have multiple ws mock server? How can it identify between them?

core/sdk/src/framework/ripple_contract.rs
satlead: Current impl would work for mocking a single websocket but we can't probably combine this with a PubSub Extension mock(which is on the horizon). Main has to have a way to detect the right extension(and processor) to send the request to.

adamdama: Sounds like a good idea. But maybe `mock_server` would be the noun?

adamdama: Done.

device/mock_device/src/mock_device_ffi.rs
adamdama: Not sure I understand the question here. I suppose the extension could start multiple websocket servers if needed but its not setup to at the moment. They would all try listening on the gateway in the Platform Parameters anyway so would fail to start.

core/sdk/src/framework/ripple_contract.rs
```diff
@@ -175,6 +180,7 @@ impl RippleContract {
         match self {
             Self::Storage(adj) => Some(adj.as_string()),
             Self::Session(adj) => Some(adj.as_string()),
+            Self::MockServer(adj) => Some(adj.as_string()),
```
adamdama, line 183: Adding all of these adjective entries caught me out at first. I wonder if there is a way to get the compiler to tell use that we have not updated all the required functions.

satlead (approved):

core/sdk/src/extn/extn_id.rs
```diff
@@ -140,6 +147,37 @@ pub struct ExtnId {
...
+        if let Ok(str) = String::deserialize(deserializer) {
+            if let Ok(id) = ExtnId::try_from(str) {
+                return Ok(id);
+            }
+        }
+        Err(serde::de::Error::unknown_variant("unknown", &["unknown"]))
+    }
+}
+
+impl PartialEq for ExtnId {
```
Vinodsathyaseelan, line 173: Why do we need this ? Is there any difference compared to the default PartialEq implementation ?

ripple
```diff
@@ -83,19 +83,40 @@ case ${1} in
...
-        sed -i "" "s@\"library\": \"/etc/firebolt-app-library.json\"@\"library\": \"~/.ripple/firebolt-app-library.json\"@" ~/.ripple/firebolt-device-manifest.json
+        sed -i "" "s@\"library\": \"/etc/firebolt-app-library.json\"@\"library\": \"$HOME/.ripple/firebolt-app-library.json\"@" ~/.ripple/firebolt-device-manifest.json
 
         echo "All Done!"
     ;;
     "run")
         cargo build --features local_dev
         THUNDER_HOST=${2} cargo run --features local_dev core/main 
     ;;
+    "run-mock")
```
Vinodsathyaseelan, line 99: Looks like some of options used in sed command works only on Mac. Need to make it compatible for Linux.

device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
```diff
@@ -994,6 +994,7 @@ impl ThunderDeviceInfoRequestProcessor {
                         params: None,
                     })
                     .await;
+                // FIXME: if the thunder plugin does not respond then we panic on the unwrap here. This would be a problem if the Thunder System plugin was not loaded
```
Vinodsathyaseelan, line 997: Can we fix this ?

core/sdk/src/extn/extn_id.rs
satlead: ExtnId will be used in the future for verifying if the request is coming from a specific Extension to return it back. PartialEq helps in validating the ExtnIds

device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
satlead: yes this is old code i fixed it

ripple
satlead: probably a separate effort

core/sdk/src/extn/extn_id.rs
Vinodsathyaseelan: Yeah, my question was more about having a custom implementation here instead of using the default one.  Now we need to maintain this custom implementation. For instance, it will break if we add an additional member field to ExtnId Struct and forget to update this custom function with additional equality check.

ripple
Vinodsathyaseelan: I was worried about the impact of using this on CI Pipeline. It's ok we we are not going to use this script on CI System.

satlead: yes this is for developer runs

core/sdk/src/extn/extn_id.rs
satlead: Agreed removed the PartialEq Implementation given all the fields now have partial eq capability

Vinodsathyaseelan (approved):

PR #242, opened 2023-09-21
 feat: Ripple should override the source field in the intent for Discovery.launch

Vinodsathyaseelan:
## What

The source field in the intent passed from an app calling Discovery.launch should not be settable by the calling application. RPPL-1004

## Why

Ripple should set the source to the calling application.
The source should be a firebolt application xrn, "xrn:firebolt:application:{appId}".

## How

## Test

## Checklist


satlead (approved):
lgtm

Swati-Chand (approved):

PR #245, opened 2023-09-22
Fixed session behavior when an app is launched again while still unloading

pahearn73:
## What

Fixes an issue where if an app that is actively unloading is re-launched, the old session ID is returned.

## Why

This behavior would result in attempts to use the old, invalid session ID.

## How

Ensure a new session ID is generated if an app launch occurs while actively unloading.

## Test

Testable by:

1. Setting lifecycle.appFinishedTimeoutMs in manifest to a larger value to delay firing of the timer.
2. Calling lifecyclemanagement.session and observe returned session IDs.
3. Calling lifecyclemanagement.setState(unloading).
4. Calling lifecyclemanagement.session and observe returned session IDs to confirm changed.
5. Use returned session ID to confirm websocket connection can be initiated.


core/main/src/bootstrap/start_fbgateway_step.rs
```diff
@@ -50,6 +50,9 @@ impl FireboltGatewayStep {
         let _ = methods.merge(DeviceRPCProvider::provide_with_alias(state.clone()));
         let _ = methods.merge(WifiRPCProvider::provide_with_alias(state.clone()));
         let _ = methods.merge(LifecycleRippleProvider::provide_with_alias(state.clone()));
+        let _ = methods.merge(LifecycleManagementProvider::provide_with_alias(
```
pahearn73, line 53: Missing, needed to pull in LifecycleManagement RPCs.

core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -843,13 +848,7 @@ impl DelegatedLauncherHandler {
             self.timer_map.insert(app_id.to_string(), unloading_timer);
         }
 
-        let timer_ms = self
```
pahearn73, line 846: Unnecessary, already done in lines above and in a separate thread.

core/main/src/bootstrap/start_fbgateway_step.rs
Vinodsathyaseelan: I was able to make `lifecyclemanagement.session() `call in gateway mode (no launcher).  @pahearn73 : Is your change applicable only in launcher mode ?

satlead: Lifecyclemanagement Provider is only used if we don't have an internal launcher please check [this](https://github.com/rdkcentral/Ripple/blob/7ee4a795ce5ce1ca111ff30710616a5f83ee9a5c/core/main/src/bootstrap/start_fbgateway_step.rs#L76) 

If the `firebolt-extn-manifest.json` contains a launcher extension it is in gateway-with-launcher if it doesn't contain the launcher extension by default it will be in gateway mode which needs the lifecyclemanagement provider.

satlead (approved):

Nv-Delle1416 (dismissed):

satlead (approved):

Nv-Delle1416 (approved):

PR #253, opened 2023-09-27
fix: device serial number in metric events is incorrect

maggie98choy:
## What

Device serial number in metric event is not populated correctly

## Why

Device mac address is set to serial number in metric events

## How

Added new rdk call to get serial number and then set its value to serial number in metric events

## Test

Validate serial number in metric event is populated with correct serial number value

## Checklist


satlead (approved):
Looks good @maggie98choy

Nv-Delle1416 (approved):

PR #259, opened 2023-10-02
feat: Do not persist a user grant if the granted return value from provider is null

bsenth200:
## What

Do not persist a user grant if the granted return value from the provider is null

## Why

A user can dismiss a user grant prompt, this should result in them not saying yes or no to the grant, but actually deferring it.

## How

Return Ungranted in an error response when the user dismisses the grant and does not persist the grant.

## Test
1. Subscribe to acknowledgechallenge.onRequestChallenge:
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "acknowledgechallenge.onRequestChallenge",
  "params": {
    "listen": true
  }
}

2. Make device.model request:
{
	"jsonrpc": "2.0",
	"id": 1,
	"method": "device.model",
	"params": {}
}

The following event will be emitted:
 {
    "id": 1,
    "jsonrpc": "2.0",
    "result": {
        "correlationId": "4fb8a921-c374-4b95-9a6f-f395e453e421",
        "parameters": {
            "capability": "xrn:firebolt:capability:device:model",
            "requestor": {
                "id": "testApp",
                "name": ""
            }
        }
    }
}

4. Now send challenge response: granted => null/false/true in acknowledgechallenge.challengeResponse request:
{
  "jsonrpc": "2.0",
  "id": 3,
  "method": "acknowledgechallenge.challengeResponse",
  "params": {
    "correlationId": <CID>,
    "result": {
      "granted": null
    }
  }
}

Expected response: You should now see device.model response based on the granted value in acknowledgechallenge.challengeResponse.
1. For granted true:
{
    "jsonrpc": "2.0",
    "result": "XiOne-SCB",
    "id": 1
}
2. For granted false:
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -40300,
        "message": "The user denied access to xrn:firebolt:capability:usergrant:acknowledgechallenge"
    }
}
2. For granted null:
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -40300,
        "message": "The user did not make a grant decision for xrn:firebolt:capability:usergrant:acknowledgechallenge"
    }
}
The user should see acknowledge again when the API is invoked next time, once after declining the grant.

## Checklist


satlead (approved):
lgtm

Swati-Chand (approved):

PR #261, opened 2023-10-03
Permission aliases

SKumarMetro:
## What

Permission alias

## Why

The Sky apps have a different set of appIds for the same app.

## How

By adding aliases

## Test
Tested using the following
{
    "jsonrpc": "2.0",
    "method": "lifecyclemanagement.session",
    "params": {
        "session": {
            "app": {
                "id": "test-refui",
## Checklist


core/main/src/state/cap/permitted_state.rs
```diff
@@ -121,11 +122,24 @@ fn get_permissions_path(saved_dir: String) -> String {
...
+        }
+    }
+
     pub async fn fetch_and_store(state: &PlatformState, app_id: &str) -> RippleResponse {
+        let app_id_alias = Self::get_distributor_alias_for_app_id(state, app_id);
         if state
             .cap_state
             .permitted_state
-            .get_app_permissions(app_id)
+            .get_app_permissions(app_id_alias.as_str())
```
Vinodsathyaseelan, line 142: cap_state should store the permission with original app_id, not with the distributor alias. So the key should be app_id here.

SKumarMetro: Replaced as suggested

Vinodsathyaseelan (dismissed):

B-Larsen (approved):

Vinodsathyaseelan (approved):

dhillomk (approved):

PR #262, opened 2023-10-03
Token support removed

joseinweb:
## What

Token support from generic devices

## Why

Without a vendor, we can't support authorization

## How

Remove the capability of the device.

## Test

Any token for token should be denied

## Checklist


satlead:
Can we merge this into https://github.com/rdkcentral/Ripple/pull/287

satlead:
Added as part of https://github.com/rdkcentral/Ripple/pull/287

PR #266, opened 2023-10-05
Usergrant request cap req cap info

Karthick-Somasundaresan:
## What

Added support for usergrants.request
## Why

This Firebolt API needs to be added to support newer version
## How

We have added new API to support this
## Test

By calling usergrants.request it should get the grants from the user according to the usergrant policy.
## Checklist


core/launcher/src/manager/app_launcher.rs
```diff
@@ -608,9 +608,13 @@ impl AppLauncher {
             }
         }
 
-        let modified_url = Self::get_modified_url(&manifest, &sessionid);
-        debug!("modified url with sessionid : {:?}", modified_url);
-        Ok(modified_url)
+        if !sessionid.is_empty() {
```
satlead, line 611: needs merging with latest code.

core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -603,7 +604,12 @@ impl DelegatedLauncherHandler {
                     false
                 }
             })
-            .filter(|perm| ps.cap_state.generic.check_all(&vec![perm.clone()]).is_ok())
```
satlead, line 606: why only supported is checked here?

core/sdk/src/api/firebolt/fb_capabilities.rs
```diff
@@ -84,6 +111,23 @@ impl FireboltCap {
         None
     }
 
+    pub fn parse_long(cap: String) -> Option<FireboltCap> {
+        let pattern = r"^xrn:firebolt:capability:([a-z0-9\\-]+)((:[a-z0-9\\-]+)?)$";
```
satlead, line 115: there is an existing parses method can we update that?

core/launcher/src/manager/app_launcher.rs
Karthick-Somasundaresan: Yes, the branch was behind the main. Now Merged it.

core/main/src/service/apps/delegated_launcher_handler.rs
Karthick-Somasundaresan: This is to get the user's consent before launching the app. It is not going to use the cap immediately. So it is enough to check if the cap is supported. Only at the time of invocation, we have check if it is available.

core/sdk/src/api/firebolt/fb_capabilities.rs
Karthick-Somasundaresan: Updated

satlead (approved):

maggie98choy (approved):

core/main/src/firebolt/handlers/capabilities_rpc.rs
```diff
@@ -222,15 +236,22 @@ impl CapabilityServer for CapabilityImpl {
         ctx: CallContext,
         grants: CapRequestRpcRequest,
     ) -> RpcResult<Vec<CapabilityInfo>> {
-        let req_list: Vec<FireboltPermission> = grants.clone().into();
-        let permitted_result =
-            PermissionHandler::check_permitted(&self.state, &ctx.app_id, &req_list).await;
+        let fb_perms: Vec<FireboltPermission> = grants.clone().into();
```
satlead, line 239: Ref

PR #277, opened 2023-10-11
fix: Serializers for SecureStorage and Language

satlead:
## What
1. Need validations for SecureStorage api(s) w.r.t Key
2. Need validations for Localization.Language api set properties
3. Need validations for Discovery watchedInfo and EntitlementsData

## Why

To enforce the firebolt schemas specified in the OpenRPC.

## How

By creating custom serializers which make sure that the incoming data is fully compliant to the RegEx specification.
RegEx used in the code matches the definition in the Open RPC.

## Test
Trigger calls for SecureStorage like Get,Set and Remove to validate the key specification.
Trigger Firebolt set call for Localization.language
Trigger Firebolt request for Discovery.contentAccess

## Checklist

### DIFF
--- a/core/sdk/src/api/device/device_request.rs
+++ b/core/sdk/src/api/device/device_request.rs
@@ -19,6 +19,7 @@ use crate::{
     api::firebolt::fb_openrpc::FireboltSemanticVersion,
     extn::extn_client_message::{ExtnEvent, ExtnPayload, ExtnPayloadProvider},
     framework::ripple_contract::RippleContract,
+    utils::serde_utils::language_code_serde,
 };
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
@@ -197,12 +198,15 @@ pub struct OnInternetConnectedRequest {
 
 #[derive(Debug, Serialize, Deserialize, Default, Clone)]
 pub struct LanguageProperty {
-    //#[serde(with = "language_code_serde")]
+    #[serde(with = "language_code_serde")]
     pub value: String,
 }
 
 #[derive(Debug, Serialize, Deserialize, Default, Clone)]
 pub struct TimezoneProperty {
+    // Original Regex in the Firebolt Timezone openrpc spec seems to be allowing
+    // even blank strings so the below test case is failing leaving it here
+    // so we can get a future resolution
     //#[serde(with = "timezone_serde")]
     pub value: String,
 }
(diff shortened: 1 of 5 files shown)


core/sdk/src/api/device/device_request.rs
```diff
@@ -197,12 +198,15 @@ pub struct OnInternetConnectedRequest {
...
+    #[serde(with = "language_code_serde")]
     pub value: String,
 }
 
 #[derive(Debug, Serialize, Deserialize, Default, Clone)]
 pub struct TimezoneProperty {
+    // Original Regex in the Firebolt Timezone openrpc spec seems to be allowing
+    // even blank strings so the below test case is failing leaving it here
+    // so we can get a future resolution
     //#[serde(with = "timezone_serde")]
```
Vinodsathyaseelan, line 210: So we are not doing strict timezone format validation here. Is the test cases still failing?

core/sdk/src/api/firebolt/fb_discovery.rs
```diff
@@ -61,10 +63,19 @@ impl LaunchRequest {
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct EntitlementData {
+    #[serde(deserialize_with = "valid_string_deserializer")]
     pub entitlement_id: String,
```
Vinodsathyaseelan, line 67: Is empty string not a valid input value for `entitlement_id` ?  Content Access is mainly a pass through feature. 
What if one of the distributors would like  to collect empty string as `entitlement_id` for some specific reason.

core/sdk/src/api/device/device_request.rs
satlead: Have used the Regex in the openrpc which is used in `serde_utils` it allows a blank timezone to be passed incidentally through the unit test. For now, let this code sit here while we fix the spec eventually we will uncomment it and have it working

core/sdk/src/api/firebolt/fb_discovery.rs
satlead: Am not sure too feels odd to not have an id given that start_time and end_time are optional too.

satlead: will revert this change

Vinodsathyaseelan (approved):
LGTM

Vinodsathyaseelan (approved):

Vinodsathyaseelan (approved):

Nv-Delle1416 (approved):

PR #279, opened 2023-10-17
feat: Pubsub Framework

Karthick-Somasundaresan:
## What

## Why

Why are these changes needed?

## How

This PR introduces structures that enable extensions to send and receive Pubsub-related requests, responses, and events 

## Test

I've tested it using test extensions

## Checklist


satlead (approved):

dhillomk (approved):

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -236,21 +236,36 @@ impl ExtnClient {
...
-                    if let Some(context) = RippleContext::is_ripple_context(&message.payload) {
-                        trace!(
-                            "Received ripple context in {} message: {:?}",
-                            self.sender.get_cap().to_string(),
-                            message
-                        );
-                        {
-                            let mut ripple_context = self.ripple_context.write().unwrap();
-                            ripple_context.deep_copy(context);
+                if is_main // This part of code is for the main ExntClient to handle
```
satlead, line 239: This new section of code handles the below expectation for Main's extension client to forward the event between one extension to the other.

Authorization to check whether a given Extension has the authority to send the event to the other extension is defined in the manifest with `ExtnSender forward_event` checks for the contract permissions. So this code is basically a pass through from and extension to main and then to the actual extension.

We will implement a separate change to just emit events from extension to extension without needing to go through main similar to how we handle the response for requests.

satlead (approved):

PR #283, opened 2023-10-19
feat: check policy even if there is existing grant state

maggie98choy:
## What

Ripple should check policy if existing grant state capabilities exist in grant policy

## Why

If policy no longer exists we should not denied a request that previous rejected by user

## How

If the policy no longer exists then delete whatever is in the grant state for that capability/role.

## Test

Have a grant policy for a capability with a lifespan of forever
Initiate the grant by trying to use the capability
Have the user reject the grant
Use the capability again, it should fail (expected)
Edit the device manifest, remove the grant policy for that capability
Reboot the device
Use the capability again

## Checklist


core/main/src/service/user_grants.rs
```diff
@@ -121,6 +121,102 @@ impl GrantState {
...
+        debug!(
+            "sync: grant entries to be removed {:#?}",
+            grant_entries_to_remove
+        );
+        grant_entries_to_remove
+    }
+
+    async fn force_delete_user_grant_from_all_sources(
+        platform_state: &PlatformState,
+        app_id: String,
```
Vinodsathyaseelan, line 175: You may change `app_id: &str` to avoid String cloning,

core/main/src/service/user_grants.rs
```diff
@@ -121,6 +121,102 @@ impl GrantState {
...
+        // Delete app grant in local storage and grant state by app id
+        let grant_entry_to_remove =
+            Self::force_delete_user_grant(platform_state, app_id.clone(), entry);
+
+        if let Some(gc_opt) = grant_entry_to_remove {
+            // Remove grant entry from cloud
+            let mut gc = gc_opt;
+            gc.status = None;
+
+            GrantPolicyEnforcer::send_usergrants_for_cloud_storage(
```
Vinodsathyaseelan, line 187: Do we need to send for all cases or only for when storage type is set to 'sync' ?

Vinodsathyaseelan (approved):

Swati-Chand (approved):

PR #285, opened 2023-10-24
feat: Discovery lifecycle issues

satlead:
## What

During the certification run the were following issues identified 
1. Lifecycle history showcases Apps going to the background from inactivity
2. Entity Type is not a required field in the OpenRPC spec yet ripple was failing without passing through
3. The action field was incorrectly added to the Tune Intent
4. Navigational Intent was setup with an untagged serde macro making it unusable for launcher extension.

## Why
We need Ripple certified for Lifecycle and Discovery.launch modules.

## How
1. Added condition to stop app going into background if its already in Foreground
2. Fix serde errors in Navigational Intent
3. Added a new internal Navigation Intent for IEC cases between Launcher extension and delegated app launcher

## Test

1. Follow the procedures set for Firebolt Certification modules specifically discovery.launch and lifecycle
2. Added unit test for reference

## Checklist


core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -95,6 +95,8 @@ pub struct App {
 #[derive(Debug, Clone, Default)]
 pub struct AppManagerState {
     apps: Arc<RwLock<HashMap<String, App>>>,
+    // Very useful for internal launcher where the intent might get untagged
+    intents: Arc<RwLock<HashMap<String, NavigationIntent>>>,
```
pahearn73, line 99: Can't you add an `intent` field to `App` so tat you can just have the one existing `apps` object?

core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -302,6 +325,15 @@ impl DelegatedLauncherHandler {
         // TODO: Add metrics helper for app loading
 
         let app_id = session.app.id.clone();
+
+        if self.platform_state.has_internal_launcher() {
+            // Specifically for internal launcher untagged navigation intent will probably not match the original cold launch usecase
```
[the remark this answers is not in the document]
satlead: Yes these issues were found during certification mode. This problem is only applicable for soft launches through Launcher extension eg

Lets say intent = x(s) s for strict/l for loose

Without launcher
lcm.session(AppSession(Intent) -> dal(intent) [stored and retrieved properly]

The current problem with the launcher
app(discovery.launch(x(s))) -> discovery_rpc(x(s)) -> dal(x(s)) -> launcher_extension(intent not serializable due to untagged so default) -> dal(BrowserSession(default intent)

Fix
app(discovery.launch(x(s))) -> discovery_rpc(x(s)) -> dal(x(s)) (keep the intent)-> launcher_extension(internal_navigation_intent) -> dal(BrowserSession(some_intent) but we use the stored intent

core/main/src/service/apps/delegated_launcher_handler.rs
satlead: App gets created using the Session call intents is stored temporarily till that call arrives and replace the intent

core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -192,12 +204,20 @@ impl DelegatedLauncherHandler {
...
                 }
                 AppMethod::Launch(launch_request) => {
+                    if self.platform_state.has_internal_launcher() {
+                        // When using internal launcher extension the NavigationIntent structure will get untagged we will use the original
+                        // intent in these cases to avoid loss of data
+                        self.platform_state.app_manager_state.store_intent(
+                            &launch_request.app_id,
+                            launch_request.get_intent().clone(),
+                        );
+                    }
```
pahearn73, line 214: Does this mean that even when building the equivalent of a "gateway_with_launcher" type ripple (e.g. for FC-style builds), we're including and building `DelegatedLauncherHandler`?

satlead: yes DelegatedLauncherHandler is bundled by default. Launcher extension is included or not depending on the distributor

pahearn73 (approved):

B-Larsen (approved):

PR #288, opened 2023-10-25
fix: capabilities info for app ignore

satlead:
## What

Add support for ignore rules for capabilities.info

## Why

Ignore rules defined in the manifest should be accounted for capabilities.info.

## How

Adding the support to ignore the checks for app ids mentioned in the manifest

## Test

1. Setup the device manifest with the app_ignore_rules for the Exclusory field
2. Call capabilities.info from the ignored app
3. Validate the results

## Checklist


Vinodsathyaseelan (approved):

bobra200 (commented):
Unit tests?

bobra200 (approved):

Nv-Delle1416 (approved):

PR #293, opened 2023-10-27
Support non-negotiable public firebolt apis

maggie98choy:
## What

Add support for non-negotiable public firebolt APIs

## Why

In the firebolt spec manifest a method can be marked as non-negotiable. For these APIs, we do not enforce any permission checking of capabilities.

## How

Add a check for non-negotiable public firebolt APIs and do not enforce any permission checking of capabilities.

## Test

Use a capability that a testing app is lack of. Update the capability in firebolt openRPC to be non-negotiable and make a request of the firebolt API that use that capability.

## Checklist


core/main/src/state/cap/generic_cap_state.rs
```diff
@@ -120,4 +122,34 @@ impl GenericCapState {
         self.check_supported(permissions)?;
         self.check_available(permissions)
     }
+
+    pub fn clear_non_negotiable_permission(
```
Vinodsathyaseelan, line 126: Can you confirm if this function removes non-negotiable permission from the input permissions list ?

Vinodsathyaseelan (changes requested):
.

core/main/src/firebolt/firebolt_gatekeeper.rs
```diff
@@ -142,8 +151,6 @@ impl FireboltGatekeeper {
                         trace!("check_with_roles for method ({}) succeded", request.method);
                     }
                 }
-            } else {
-                debug!("Role/Capability is cleared based on non-negotiable policy");
             }
         } else {
             // Couldnt find any capabilities for the method
```
Vinodsathyaseelan: I think it's not correct to return error with DenyReason::Unsupported for the use case where all the required permission(s) are filtered out due to non-negotiable policy.  In this case, the permission check should return Ok(),
indicating that the required capabilities are permitted.

Vinodsathyaseelan (approved):

Swati-Chand (approved):

PR #296, opened 2023-11-01
feat: Changes for 1.0.0

satlead:
## What
Removes /opt and other overrides for manifest making them compile time build configuration
Upgrades the developer run scripts with right build flags
Upgrades version numbers

## Why

First step to make Ripple production ready.

## How

Using Rust configuration flags

## Test
Execute ./ripple run <ip_address> 

## Checklist

### DIFF
--- a/core/main/src/bootstrap/manifest/device.rs
+++ b/core/main/src/bootstrap/manifest/device.rs
@@ -35,10 +35,12 @@ impl LoadDeviceManifestStep {
 type DeviceManifestLoader = Vec<fn() -> Result<(String, DeviceManifest), RippleError>>;
 
 fn try_manifest_files() -> Result<DeviceManifest, RippleError> {
-    let dm_arr: DeviceManifestLoader = if cfg!(test) {
+    let dm_arr: DeviceManifestLoader = if cfg!(any(feature = "local_dev", feature = "pre_prod")) {
+        vec![load_from_env, load_from_home, load_from_opt, load_from_etc]
+    } else if cfg!(test) {
         vec![load_from_env]
     } else {
-        vec![load_from_env, load_from_home, load_from_opt, load_from_etc]
+        vec![load_from_etc]
     };
 
     for dm_provider in dm_arr {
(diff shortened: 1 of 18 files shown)


core/main/src/bootstrap/manifest/device.rs
```diff
@@ -35,10 +35,12 @@ impl LoadDeviceManifestStep {
...
 
 fn try_manifest_files() -> Result<DeviceManifest, RippleError> {
-    let dm_arr: DeviceManifestLoader = if cfg!(test) {
+    let dm_arr: DeviceManifestLoader = if cfg!(any(feature = "local_dev", feature = "pre_prod")) {
+        vec![load_from_env, load_from_home, load_from_opt, load_from_etc]
+    } else if cfg!(test) {
         vec![load_from_env]
     } else {
-        vec![load_from_env, load_from_home, load_from_opt, load_from_etc]
+        vec![load_from_etc]
```
pahearn73, line 43: Will this prevent the ability to debug prod releases using overridden manifests if necessary?

core/main/src/bootstrap/manifest/extn.rs
```diff
@@ -30,10 +30,12 @@ impl LoadExtnManifestStep {
 type ExtnManifestLoader = Vec<fn() -> Result<(String, ExtnManifest), RippleError>>;
 
 fn try_manifest_files() -> Result<ExtnManifest, RippleError> {
-    let dm_arr: ExtnManifestLoader = if cfg!(test) {
+    let dm_arr: ExtnManifestLoader = if cfg!(any(feature = "local_dev", feature = "pre_prod")) {
```
pahearn73, line 33: What will be the expected differences between `local_dev` and `pre_prod`?

core/main/src/bootstrap/manifest/device.rs
satlead: Debug releases probably can be enabled with Rust log of debug or trace for triaging. Having this Prod builds opens up concerns with security this reduces the attack radius on a prod builds

core/main/src/bootstrap/manifest/extn.rs
satlead: Probably one for the future where we can configure local dev shell scripts and pre-prod environments differently.
`pre_prod` can be used by ci cd pipelines to generate builds.

We can configure other items for local_dev alone to make it customizable felt its a good tangent

pahearn73 (approved):

kevinshahfws (approved):

PR #299, opened 2023-11-04
feat: handle missing provider case for evaluateAt=activeSession

maggie98choy:
## What

Add checking on required provider is available for evaluateAt="activeSession" case

## Why

An app requires an user grant to start an app session but nothing provide the granting capability

## How

Ripple should not return transitionPending=true because the session transition will never be complete because it has already failed

## Test
- Starts an app session when `acknowledgeChallenge.onRequestChallenge` is not registered. Validate "transitionPending": false` is returned from response.
- Starts an app session when `acknowledgeChallenge.onRequestChallenge` has been registered. Validate "transitionPending": true` is returned from response.

## Checklist


core/main/src/service/user_grants.rs
```diff
@@ -1610,6 +1667,16 @@ mod tests {
             };
             let ctx = runtime.call_context;
             let mut platform_state = runtime.platform_state;
+            let caps = vec![
```
satlead, line 1670: Isnt this per provider availability

satlead: Ok this is just for unit test setup

satlead (approved):

Vinodsathyaseelan (approved):

Swati-Chand (approved):

PR #300, opened 2023-11-07
Lifespan once fix

Karthick-Somasundaresan:
## What

When the life scope of grant policy is set to once. the grants should not be saved. But when the user grants are granted via usergrants.grant/deny, the grant is getting saved in the file. 

## Why

These changes are needed to avoid duplicate grants when requested due to the fact that they are saved.

## How

This change ensures that before saving the grant's it checks the policy lifespan and writes to the file only when lifespan is not once.

## Test

Yes this has been tested by calling usergrants.grant/deny for a capability with policy and lifespan set to once and I have checked the saved grant entries are not created in the file when the lifespan is set to once.

## Checklist


satlead (approved):
It would be ideal to add unit tests to changes like these.
Good to go for QA

Swati-Chand (approved):

PR #301, opened 2023-11-07
feat: Self contained extension provider

satlead:
## What

Provides SDK support for using ExtnId based providers

## Why

To avoid adding new contracts for self contained extensions

## How

By using ExtnId to forward the request to the intended processors

## Test

## Checklist


device/mock_device/src/lib.rs
```diff
@@ -20,8 +20,8 @@ pub mod mock_data;
 pub mod mock_device_controller;
 pub mod mock_device_ffi;
 pub mod mock_device_processor;
+pub mod mock_server;
 pub mod mock_web_socket_server;
-pub mod utils;
-
 #[cfg(test)]
 pub(crate) mod test_utils;
+pub mod utils;
```
adamdama, line 27: Why has this moved? The functions in here are needed to start the extension.

satlead: accidental during merge will fix

PR #308, opened 2023-11-09
Fix BI Metrics app ver and os ver incorrect

maggie98choy:
## What

Set BI metrics app_ver field to ripple version and BI metrics os_ver field to firmware version

## Why

There are not being set correctly

## How

`app_ver` field is set to ripple version. 
`os_ver` field is set to the value from getSystemVersions thunder call.

## Checklist


core/sdk/Cargo.toml
```diff
@@ -20,6 +20,7 @@ name = "ripple_sdk"
 version = "1.0.0"
 edition = "2021"
 repository = "https://github.com/rdkcentral/Ripple"
+build = "build.rs"
```
satlead, line 23: Can we take this from main instead of SDK

core/sdk/build.rs
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+extern crate vergen;
+use vergen::{vergen, ConstantsFlags};
+
+fn main() {
```
satlead, line 21: SDK should not need a main

core/sdk/src/api/firebolt/fb_metrics.rs
```diff
@@ -47,7 +48,7 @@ impl From<CallContext> for BehavioralMetricContext {
     fn from(call_context: CallContext) -> Self {
         BehavioralMetricContext {
             app_id: call_context.app_id.clone(),
-            app_version: String::from("app.version.not.implemented"),
+            app_version: SEMVER.to_string(),
```
satlead, line 51: Main can replace this on the metrics context

core/sdk/build.rs
brendanobra: This main for build, not main (main) (build.rs, not main.rs). This is required to get the version to be generated correctly.

brendanobra (approved):

satlead (approved):

B-Larsen (approved):

PR #310, opened 2023-11-10
feat: Performance Improvements Iec async channel

satlead:
## What
Current implementation of Ripple has issues related to
1. RPC Call Latency
2. CPU Usage 

## Why

Root cause for this problem is Usage of Crossbeam channel in IEC between Main and Extensions
1. Using a synchronous cross-beam channel in the IEC. Because this is synchronous we manually add an interval and verify if there are messages in the IEC.
2. This also adds a random delay of 5 milli seconds between each packet between extensions and main.
3. Even when Idle this process keeps happening to check if there are packets in the IEC for main and each of the extensions.

## How

This Pr updates our IEC engine from Crossbeam channel to Async Channel crate. This new crate supports asynchronous operations without having the need for interval based checks.

## Test

How was this tested
1. Build Ripple OSS and run it as a system d process.
2. Having ripple idle issue a top command on the device and verify the memory usage
3. Check the RSS Private and Shared usage

## Checklist


core/sdk/src/extn/client/extn_client.rs
```diff
@@ -21,8 +21,9 @@ use std::{
     time::Duration,
 };
 
+use async_channel::{bounded, Receiver as CReceiver, Sender as CSender, TryRecvError};
 use chrono::Utc;
-use crossbeam::channel::{bounded, Receiver as CReceiver, Sender as CSender, TryRecvError};
+//use async_channel::{bounded, Receiver as CReceiver, Sender as CSender, TryRecvError};
```
adamdama, line 26: Not sure you need this line.

satlead: Yes will clean up the comment

adamdama (approved):
This is amazing! The main loop is so much smaller and the mode of operation now is much more elegant. Awesome work here.

mhughesacn:
Hi @satlead : There are JPG files in this pull req.  RDK CMF is nervous about binary content in hosted git repositories.  This can lead to performance problems, longer clone times etc.  Please try to keep this material to a minimum without large images, or large numbers of images.  Thank you!
- Martin Hughes, RDK CMF Compliance Team

satlead:
> Hi @satlead : There are JPG files in this pull req. RDK CMF is nervous about binary content in hosted git repositories. This can lead to performance problems, longer clone times etc. Please try to keep this material to a minimum without large images, or large numbers of images. Thank you!
> 
> * Martin Hughes, RDK CMF Compliance Team

Sure thing i will clean this up

satlead:
Removed all the images @mhughesacn

mhughesacn:
Many thanks Satishkumar.

core/main/src/main.rs
```diff
@@ -30,7 +30,7 @@ pub mod state;
 pub mod utils;
 include!(concat!(env!("OUT_DIR"), "/version.rs"));
 
-#[tokio::main]
+#[tokio::main(worker_threads = 2)]
```
Vinodsathyaseelan, line 33: Is there any specific reason for configuring worker_threads count to 2 ?

satlead: Based on analysis with the new changes 2 worker threads seem more than enough for handling the load of Ripple. Logic behind this is we have 2 websockets one for 3474 and 3473. So the worker threads will be used by tokio runtime to sequester the next tokio task in one of the workers depending on the availability. 2 would be more optimal than over allocation of 4

Vinodsathyaseelan (approved):

PR #313, opened 2023-11-11
feat: Support settings get and sub with alias

satlead:
## What

Add Settings contract support for Get and Subscribe using Aliases

## Why

Distributors can use these changes to migrate their existing settings values using aliases
For eg VoiceGuidance can be stored as tts(text to speech) enabled by the distributor.
```
let keys = vec![SettingKey::VoiceGuidanceEnabled]
let alias = HashMap::new();
alias.insert(SettingKey::VoiceGuidanceEnabled.to_string(), "tts".to_owned())

SettingsRequest::get(SettingsRequestParam(ctx, keys, Some(alias))
```

Response would be 
```
... 
{
   "tts": {
         "enabled": true
    }
}
...
```
## How

Provide support for aliasing the settings recognized by Ripple to strings which can be aliased by the distributor.

## Test

Create an alias in an extension and send the request to Ripple main as a Get request. Extensions can also use the subscribe method to listen to changes in settings using aliases.

## Checklist


core/main/src/firebolt/handlers/privacy_rpc.rs
```diff
@@ -298,35 +299,68 @@ pub struct PrivacyImpl {
...
-        AppEvents::add_listener_with_context(
-            &self.state,
-            event_name.to_owned(),
-            ctx,
-            ListenRequest {
-                listen: request.listen,
-            },
-            event_context,
-        );
+        if dec.is_some() {
```
Vinodsathyaseelan, line 323: map() function can be used here to make it more idiomatic.

Vinodsathyaseelan (approved):

Vinodsathyaseelan (approved):

Nv-Delle1416 (approved):

PR #315, opened 2023-11-14
Added aarch64 workflow

StefanBossbaly:
## What
Given that the vast majority of our target devices use the aarch64 ISA, it would be beneficial to run a build and test using that architecture. Since Github runners use x64, we have to use qemu to emulate both the build and test environment. There is a signifact performance penalty when using qemu. It might be possible to do the build using the aarch64-unknown-linux-gnu toolchain on x64 and just run the unit tests in the qemu environment, but it would require the sysroot and rootfs to be binary compatible.

## Why

Difference in types (especially when using FFI functions) can cause compilation errors. This ensure that future changes must pass both x64 and aarch64 workflows inorder to be considered for a merge into master.

## How

By using qemu, we are able to confirm that Ripple can compile on aarch64, the target architecture.

## Test

Verify that the new Github action passes.

## Checklist

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -85,3 +85,35 @@ jobs:
           append: true
           path: current_coverage.md
         continue-on-error: true
+
+  build_and_test_aarch64:
+    name: Build and Test on aarch64
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Build and Test
+        uses: uraimo/run-on-arch-action@v2.5.1
+        id: build_and_test
+        with:
+          arch: aarch64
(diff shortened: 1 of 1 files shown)


brendanobra (approved):

StefanBossbaly:
@brendanobra Two things to note.

1) This workflow does take 50 mins to complete. Up to the Ripple team if that is too long to make a blocking merge.
2) This workflow will create an image of the docker container used to build and run the workflow on aarch64 and save it to this repo's packages. This speeds up the process as subsequent builds don't have to completely rebuild the image. No concerrns just something the team should be aware of. You can read more about it here: https://github.com/uraimo/run-on-arch-action#usage

satlead (approved):

PR #316, opened 2023-11-15
fix: capabilities.granted returns null if Grant state does not exist …

maggie98choy:
## What

capabilities.granted returns null if Grant state does not exist and there is NO grant policy for the capability

## Why

To align with firebolt spec.

## How

If grant state does not exist for the capability then check if grant policy exist for the capability.

## Test

validate case where grant state does not exist and there is NOT a policy for the capability

## Checklist

### DIFF
--- a/core/main/src/service/user_grants.rs
+++ b/core/main/src/service/user_grants.rs
@@ -497,11 +497,10 @@ impl GrantState {
         if let Some(status) = self.get_grant_status(app_id, permission) {
             GrantActiveState::ActiveGrant(status.into())
         } else {
-            // This flow is where the capability is not in grant state.
+            // This flow is when the capability is not in grant state.
             // Check if there is a grant policy for the capability.
             // if no, return allowed as active grant state;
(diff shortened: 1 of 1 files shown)


core/main/src/service/user_grants.rs
```diff
@@ -492,10 +492,29 @@ impl GrantState {
...
+        platform_state: Option<&PlatformState>,
     ) -> GrantActiveState {
         if let Some(status) = self.get_grant_status(app_id, permission) {
             GrantActiveState::ActiveGrant(status.into())
         } else {
+            // This flow is where the capability is not in grant state.
+            // Check if there is a grant policy for the capability.
+            // if no, return allowed as active grant state;
+            // otherwise, return pending grant.
+            // https://ccp.sys.comcast.net/browse/RPPL-914
```
satlead, line 504: Remove the link

satlead (approved):

satlead (approved):

Swati-Chand (approved):

PR #319, opened 2023-11-16
fix: Adding app title in usergrants api

Karthick-Somasundaresan:
## What

Fill the app title info in user grants apis

## Why

To be in line with the spec
## How

This adds the provision to store the app titles that were launched and use that info to fill the app title info in user-grants api 

## Test

I have made the JSONRPC call and ensured that the app Ids are added. 

## Checklist


core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -105,9 +107,94 @@ pub struct AppManagerState {
...
+        let stored_value_res = Self::restore_app_info_from_storage(storage_path);
+        if let Ok(stored_value) = stored_value_res {
+            let map_res: Result<HashMap<String, String>, _> = serde_json::from_value(stored_value);
+            map_res.unwrap_or_default()
+        } else {
+            HashMap::new()
+        }
+    }
+    fn get_storage_path(saved_dir: &str) -> String {
+        let mut path = std::path::Path::new(saved_dir).join("app_info/");
```
Vinodsathyaseelan, line 149: Please check the usage of format!(" {} / {}" ) in file path.  Instead  use `PathBuf` or `Path` to build the file url. 
Avoid using explicit file path delimiter ("/"), let `join() `take care of that.

Vinodsathyaseelan (approved):

PR #320, opened 2023-11-16
Discovery.on policy changed for share watch history

SKumarMetro:
## What

Discovery.onPolicy changed for share watch history

## Why

Why are these changes needed?

## How

## Test

## Checklist


satlead:
We are trying to evaluate the original implementation and try to reuse this change to create an association between privacy.shareWatchHistory with discovery:watched using the manifest instead of Ripple RPC handler

PR #323, opened 2023-11-17
Return resolution in order of availability: 1) Current, 2) System default, 3) Manifest default.

pahearn73:
Changed video resolution queries so that instead of returning the system default value, it first attempts to determine the current video resolution. If not available it queries the system default, and finally returns the default specified in the device manifest.


satlead (approved):

device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
```diff
@@ -820,7 +823,26 @@ impl ThunderDeviceInfoRequestProcessor {
...
+            })
+            .await;
+        if response.message.get("success").is_none()
+            || !response.message["success"].as_bool().unwrap_or_default()
+        {
+            error!("{}", response.message);
+            return Err(());
+        }
+        info!("{}", response.message);
+        let resol = response.message["resolution"].as_str().unwrap();
```
Vinodsathyaseelan, line 841: What if the response message doesn't have `resolution`  field ?  
`as_str() ` will return `None` , leads to panic!

Vinodsathyaseelan (approved):

PR #324, opened 2023-11-17
fix: Handling watched and watchNext scenarios

satlead:
## What

Cleanup Account Processor for Watched and WatchNext Api(s)

## Why

ProgressUnit for WatchNext needs to be different from the progress value of `watched_info` to allow the distributor to define the progress value.

## How

Providing a new ProgressUnit gives the distributor the ability to configure. 
## Test

## Checklist

### DIFF
--- a/core/sdk/src/api/firebolt/fb_discovery.rs
+++ b/core/sdk/src/api/firebolt/fb_discovery.rs
@@ -118,6 +118,17 @@ pub struct WatchedInfo {
     )]
     pub watched_on: Option<String>,
 }
+
+impl WatchedInfo {
+    pub fn get_progress(&self) -> ProgressUnit {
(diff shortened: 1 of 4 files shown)


core/sdk/src/api/firebolt/fb_discovery.rs
```diff
@@ -118,6 +118,17 @@ pub struct WatchedInfo {
     )]
     pub watched_on: Option<String>,
 }
+
+impl WatchedInfo {
+    pub fn get_progress(&self) -> ProgressUnit {
+        if self.progress.ge(&1.0) {
```
Vinodsathyaseelan, line 124: Is usage of operator here make it more idiomatic ?

pahearn73 (approved):

PR #326, opened 2023-11-27
Remove unnecessary clones (fixes #50)

### LINKED ISSUE #50: Add unit test coverage tooling
Add code coverage tooling to the project. We need to be able to check the test coverage of the project in the CI process at a minimum.

Options are to use Tarpaulin or grcov and xtasks as per https://blog.rng0.io/how-to-do-code-coverage-in-rust.

Must haves:
- Readable coverage output in the CI chain
- CI/CD action that compares coverage on main to the coverage on the new branch
- No ability to edit the coverage files to bypass the check

Nice to have:
- developer to run locally on ARM or x86 machine
- pre-push git hook to run code overage tooling and warn users of reduced coverage

StefanBossbaly:
## What

Removes unnecessary clones

## Why

Clones increase memory usage and increase processing time.

## How

Removes unnecessary clones by:
  1) Removing the clone() call
  2) Borrowing the value instead
  3) Defering the clone() call on the item that needs to be cloned and
     not the whole containing datastructure

## Test

cargo test

## Checklist

### DIFF
--- a/core/launcher/src/launcher_state.rs
+++ b/core/launcher/src/launcher_state.rs
@@ -59,9 +59,9 @@ impl LauncherState {
         self.extn_client.clone().request(payload).await
     }
 
-    pub async fn new(extn_client: ExtnClient) -> Result<LauncherState, RippleError> {
+    pub async fn new(mut extn_client: ExtnClient) -> Result<LauncherState, RippleError> {
         let extn_message_response: Result<ExtnMessage, RippleError> =
-            extn_client.clone().request(Config::LauncherConfig).await;
+            extn_client.request(Config::LauncherConfig).await;
         if let Ok(message) = extn_message_response {
             if let Some(config) = message.payload.extract() {
                 return Ok(LauncherState {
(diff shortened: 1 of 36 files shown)


satlead (approved):
Nice improvement on this one. Was this found by Clippy? if so are we missing lints in our ci?

adamdama (approved):
This looks good. I love a good cleanup. I second @satlead's question though it would be ace if these could be covered by the code analysis.

StefanBossbaly:
@adamdama & @satlead There is https://rust-lang.github.io/rust-clippy/master/#/redundant_clone but it will only catch _redundant_ clones, not _unnecessary_. Meaning that if an object is cloned and dropped without any actions taking place it will be able to catch that. Unfortunately it looks like cloning an object counts as an action which is why the lint was not triggering. There doesn't look like there is a lint for cloning a whole data structure (i.e. HashMap) to only move out a value. Not sure if it is possible but might be something to bring up to the clippy team.

Nv-Delle1416 (approved):

PR #328, opened 2023-11-27
fix: uncomment validate_metrics_action_type()

SKumarMetro:
## What

uncomment validate_metrics_action_type()

## Why

bug fix

## How

uncomment validate_metrics_action_type()

## Test

## Checklist


core/main/src/firebolt/handlers/metrics_rpc.rs
```diff
@@ -346,8 +346,7 @@ impl MetricsServer for MetricsImpl {
         }
     }
     async fn action(&self, ctx: CallContext, action_params: ActionParams) -> RpcResult<bool> {
-        //call it and let it blow up
-        //let _ = validate_metrics_action_type(&action_params.action_type)?;
+        let _ = validate_metrics_action_type(&action_params.action_type)?;
```
pahearn73, line 349: Should it not return the error if it's out of range?

Swati-Chand (approved):

PR #329, opened 2023-11-28
Added debugging to load_extension_library

StefanBossbaly:
## What

If a plugin fails to load, we should inform the user as to why it was not successful. Also remove unwrap() call to ensure that future code changes don't cause a panic!.

## Why

Lesser programmers (like myself) sometimes write buggy plugins that have a local sysroot and then blow up when Ripple tries to dlopen their shared objects dependencies which don't exist on the rootfs. It is helpful to output the reason why the the plugin could not be loaded.

## How

Output the error message to the log. Also remove the `unwrap()` and use `and_then()` to chain the Option processing.

## Test

My buggy plugin triggers this and gives the following output:

```
2023-11-28-20:36:45.397[DEBUG][ripple::bootstrap::extn::load_extn_metadata_step][gateway]-******************Loading /home/workuser/Ripple/target/debug/libsec_manager_extn.so************************
2023-11-28-20:36:45.397[DEBUG][ripple::bootstrap::extn::load_extn_metadata_step][gateway]-Extn not found: DlOpen { desc: "libsaclient.so.3.3.0: cannot open shared object file: No such file or directory" }
2023-11-28-20:36:45.397[WARN][ripple::bootstrap::extn::load_extn_metadata_step][gateway]-file=/home/workuser/Ripple/target/debug/libsec_manager_extn.so doesnt contain a valid extension library
```

## Checklist


brendanobra (approved):

satlead (approved):

maggie98choy:
@StefanBossbaly I updated the main branch to resolve code coverage pr check failure. Please update your branch with the latest main branch.

StefanBossbaly:
@maggie98choy Looks like the tests are passing!

Nv-Delle1416:
@StefanBossbaly please update your PR and go ahead and merge your changes to main branch

PR #350, opened 2023-12-06
Fix session id metrics is empty

maggie98choy:
## What

Assign default string to device_session_id in context

## Why

session id in metric payload is empty

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/metrics_management_rpc.rs
+++ b/core/main/src/firebolt/handlers/metrics_management_rpc.rs
@@ -15,7 +15,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 use jsonrpsee::{core::RpcResult, proc_macros::rpc, RpcModule};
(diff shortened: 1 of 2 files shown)


core/main/src/firebolt/handlers/metrics_management_rpc.rs
```diff
@@ -105,7 +107,9 @@ impl MetricsManagementServer for MetricsManagementImpl {
         for key in request.keys {
             // currently handling only one key which is deviceSessionId
             if key.as_str() == "deviceSessionId" {
-                self.state.metrics.update_session_id(None);
+                self.state
+                    .metrics
+                    .update_session_id(Some(Uuid::new_v4().to_string()));
```
Vinodsathyaseelan, line 112: As per the requirement, session_id should MetricsManagement.removeContext should revert Ripple back to its fallback value.  That value is the one created during `initialize`  not a new Uuid.

Vinodsathyaseelan (approved):

PR #351, opened 2023-12-11
Metrics updates

satlead:
## What

Adds partner Id information to metrics information
Handles the lifecycle of the app based on inactivity 

## Why
Currently apps which are launched with an intent to be inactive send invalid metric events.

## How
Representing the app events differently for apps which were launched with inactivity 

## Test
Launch apps with inactive set to true during lifecyclemanagement.session and one without. Go through the app lifecycle by launching and dismissing check logs for BehaviorMetrics

## Checklist


core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -288,7 +288,16 @@ pub struct DelegatedLauncherHandler {
 /*
 Tell lifecycle metrics which methods map to which metrics AppLifecycleStates
 */
-fn map_event(event: &AppMethod) -> Option<AppLifecycleState> {
+
+fn map_event(event: &AppMethod, inactive: bool) -> Option<AppLifecycleState> {
```
pahearn73, line 292: I'm a little confused by what "inactive" means here, it doesn't just mean the app is in the ``LifecycleState::Inactive`` state does it?

core/sdk/src/api/firebolt/fb_lifecycle.rs
```diff
@@ -36,6 +38,19 @@ pub enum LifecycleState {
     Suspended,
 }
 
+impl From<&LifecycleState> for AppLifecycleState {
```
pahearn73, line 41: If these are 1:1 mapping, why do we need `AppLifecycleState` at all , can't we just use `LifecycleState` where `AppLifecycleState` is used?

core/main/src/service/apps/delegated_launcher_handler.rs
satlead: it means it was launched as inactive so a lifecycle.ready during a launch of inactive means its actually suspended. Thats the the current change to handle apps which are launched with an intent to be suspended. Metrics would now reflect the same.

core/sdk/src/api/firebolt/fb_lifecycle.rs
satlead: AppLifecycleState is the Metrics LifecycleState they act a little differently compared to the regular app lifecycle state with apps launched with an intent of inactive: true. This change captures those use cases and makes the adjustments for those app launches.

B-Larsen (approved):

PR #357, opened 2023-12-13
fix: Making token:account as available instead of token:platform

Karthick-Somasundaresan:
## What

This PR makes token:account instead of token:platform

## Why

token:account is initially marked as unavailable but token:platform is marked available when account.session is set. 

## How

These changes will enable capturing account:session value in the ripple context.

## Test

The extension that handles account.session contract, sends RippleContextUpdateRequest::Token event, it gets stored in the ripple core. 

## Checklist


satlead (approved):
Changes look good but in the event of token:platform being not available is that gonna cause any issues for Apps which already use these API(S). cc @rdkcentral/ripple-qa

Nv-Delle1416 (approved):
You are clear to merge into 1.4.0 release

PR #361, opened 2023-12-15
fix: Grant check fails to do case-insensitive comparison of privacy settings method

Vinodsathyaseelan:
## What

When the user grants policy has an entry for a privacy setting, the property (privacy method) will be checked to validate certain conditions.

To ensure backward compatibility of Firebolt API, the module part of the method needs to be compared in a case-insensitive manner. This was missing in the current implementation

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/service/user_grants.rs
+++ b/core/main/src/service/user_grants.rs
@@ -1446,6 +1446,10 @@ impl GrantPolicyEnforcer {
         );
         let method_name = Self::get_setter_method_name(platform_state, &privacy_setting.property);
         if method_name.is_none() {
+            error!(
+                "Unable to find setter method for property: {}",
+                privacy_setting.property.as_str()
+            );
             return;
         }
         let method_name = method_name.unwrap();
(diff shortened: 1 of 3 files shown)


B-Larsen (approved):

core/main/src/service/user_grants.rs
```diff
@@ -2175,7 +2179,9 @@ mod tests {
                 result.err().unwrap(),
                 DenyReasonWithCap {
                     reason: DenyReason::GrantDenied,
-                    caps: vec![FireboltCap::Full(PIN_CHALLENGE_CAPABILITY.to_owned())]
+                    caps: vec![FireboltCap::Full(
+                        "xrn:firebolt:capability:localization:postal-code".to_owned()
```
satlead, line 2183: hardcoded to capability

satlead (approved):

core/main/src/service/user_grants.rs
Vinodsathyaseelan: Test code asserting the result.

PR #369, opened 2023-12-22
1.0.1 cp

satlead:
## What

Refer https://github.com/rdkcentral/Ripple/pull/368

## Why

Refer https://github.com/rdkcentral/Ripple/pull/368

## How

Refer https://github.com/rdkcentral/Ripple/pull/368

## Test

Refer https://github.com/rdkcentral/Ripple/pull/368

## Checklist

### DIFF
--- a/core/main/src/processor/settings_processor.rs
+++ b/core/main/src/processor/settings_processor.rs
@@ -19,6 +19,7 @@ use std::collections::HashMap;
 
 use ripple_sdk::{
     api::{
+        distributor::distributor_privacy::ContentListenRequest,
         firebolt::{
             fb_capabilities::{CapabilityRole, FireboltCap, RoleInfo},
             fb_general::ListenRequest,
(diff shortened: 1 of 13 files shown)


core/main/src/processor/settings_processor.rs
```diff
@@ -232,7 +233,10 @@ impl SettingsProcessor {
                         true,
                         &ctx,
                         EVENT_ALLOW_PERSONALIZATION_CHANGED,
-                        None,
+                        Some(ContentListenRequest {
```
brendanobra, line 236: Any way to clean this up a little? from a quick read, looks like the last key to set result "wins".  For example, if 3 keys are sent in the request to be set, the first one fails, and 2 and 3 succeed, is true to the correct return value?

satlead: Current logic is the default for resp = true if there are issues in one of the entries it becomes false but the iteration goes on.

Karthick-Somasundaresan (approved):

brendanobra (approved):

PR #373, opened 2024-01-04
Application Catalog and Management support

pahearn73:
## What

Adds general support for application catalog monitoring and updating of installed applications.

## Why

Allows distributors to define and alter the installed applications on devices through a remote catalog system.

## How

Provides a generic AppsUpdater interface for ripple, the implementation of which is determined by the particular distributor extensions installed.

## Test

Tested ripple in a local development environment using a development distribution platform instance.

## Checklist


core/main/src/bootstrap/boot.rs
```diff
@@ -64,6 +65,7 @@ pub async fn boot(state: BootstrapState) -> RippleResponse {
     execute_step(LoadExtensionsStep, &bootstrap).await?;
     execute_step(StartExtnChannelsStep, &bootstrap).await?;
     execute_step(StartAppManagerStep, &bootstrap).await?;
+    execute_step(StartAppsUpdaterStep, &bootstrap).await?;
```
satlead, line 68: We cleaned up the steps that are needed post-session availability to the context manager. The contents of this step should be added as a [method](https://github.com/rdkcentral/Ripple/blob/613d8b283f52dcce50c7b4bc0f29d3cfeca84664/core/main/src/processor/main_context_processor.rs#L128) call in Context manager

kpears201 (approved):

core/main/src/processor/main_context_processor.rs
```diff
@@ -125,47 +126,64 @@ impl MainContextProcessor {
...
+                            .await;
+                        debug!(
+                            "Received Sync response for user grants: {:?}",
+                            sync_response
+                        );
+                    }
                 }
             }
+
+            if state.supports_app_catalog() {
```
satlead, line 174: Perfect

satlead (approved):

B-Larsen (approved):

PR #377, opened 2024-01-08
Remove unwraps

StefanBossbaly:
## What

Removes ~90 `.unwrap()` calls in favor of other Rust programming paradigms that prove the state of the enum instead of generating a `panic!()` on unexpected case.

Most the the replaced code was written like so, with the conditional check to handle the unwanted variant (`None` in this case) and then the call to `unwrap()` to extract the value.

```
if entry.is_none() {
    error!(
        "on_app_state_change: app_id={} Not found",
        state_change.container_props.name
    );
    return;
}

// safe to unwrap
let app = entry.unwrap();
```

However `unwrap()` under the hood is just a `panic!` when the `None` variant is encountered like so.

```
if entry.is_none() {
    error!(
        "on_app_state_change: app_id={} Not found",
        state_change.container_props.name
    );
    return;
}

// safe to unwrap
let app = match entry {
  Some(value) => value,
  None => panic("called `Option::unwrap()` on a `None` value"),
};
```

You can simplify this code by simply handling the `None` case with a `match` statement. This is remove the panic to ensure that future code changes do not trigger that case like so:

```
let app = match entry {
  Some(value) => value,
  None => {
    error!(
        "on_app_state_change: app_id={} Not found",
        state_change.container_props.name
    );
    return;
  }
};
```

## Why

Calling `.unwrap()` is an anti-pattern in Rust (with limited exceptions). It is better to use pattern matching or other provided functions to handle the cases where the enum is not the expected variant. This also has the added benefit of reducing code size since the `panic!()` macro to handle the unexpected variant will not be generated.

## How

Removes `unwrap()` calls in favor of pattern matching and other monad functions.

## Test

`cargo build --release`, `cargo test` and inspection to prove that there is no functional difference between the code.

## Checklist

### DIFF
--- a/core/launcher/src/manager/view_manager.rs
+++ b/core/launcher/src/manager/view_manager.rs
@@ -88,11 +88,12 @@ impl ViewRequest {
 
     pub fn send_response(&self, response: ViewResponse) -> Result<(), RippleError> {
         let mut sender = self.resp_tx.write().unwrap();
-        if sender.is_some() {
-            oneshot_send_and_log(sender.take().unwrap(), response, "ViewManager response");
-            Ok(())
-        } else {
-            Err(RippleError::SenderMissing)
+        match sender.take() {
+            Some(tx) => {
+                oneshot_send_and_log(tx, response, "ViewManager response");
+                Ok(())
+            }
+            None => Err(RippleError::SenderMissing),
         }
     }
 }
--- a/core/main/src/firebolt/firebolt_gatekeeper.rs
+++ b/core/main/src/firebolt/firebolt_gatekeeper.rs
@@ -66,8 +66,8 @@ impl FireboltGatekeeper {
         let perm_based_on_spec_opt = platform_state
             .open_rpc_state
             .get_perms_for_method(method, api_surface);
-        perm_based_on_spec_opt.as_ref()?;
-        let perm_based_on_spec = perm_based_on_spec_opt.unwrap();
+
+        let perm_based_on_spec = perm_based_on_spec_opt?;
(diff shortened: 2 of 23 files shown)


core/launcher/src/manager/view_manager.rs
```diff
@@ -88,11 +88,12 @@ impl ViewRequest {
...
-            oneshot_send_and_log(sender.take().unwrap(), response, "ViewManager response");
-            Ok(())
-        } else {
-            Err(RippleError::SenderMissing)
+        match sender.take() {
+            Some(tx) => {
+                oneshot_send_and_log(tx, response, "ViewManager response");
+                Ok(())
+            }
+            None => Err(RippleError::SenderMissing),
```
pahearn73, line 96: Clippy tends to favor is_some but we can see what happens, either of these are fine.

core/main/src/firebolt/firebolt_gatekeeper.rs
```diff
@@ -66,8 +66,8 @@ impl FireboltGatekeeper {
         let perm_based_on_spec_opt = platform_state
             .open_rpc_state
             .get_perms_for_method(method, api_surface);
-        perm_based_on_spec_opt.as_ref()?;
-        let perm_based_on_spec = perm_based_on_spec_opt.unwrap();
+
+        let perm_based_on_spec = perm_based_on_spec_opt?;
```
pahearn73, line 70: Probably just get rid of `perm_based_on_spec_opt` here.

pahearn73 (approved):

PR #378, opened 2024-01-09
support for web apps.

kvfasil:
## What

Adding support for web apps.

## Why

Currently web apps are not supported.

## How

by listening to launcher events.

## Test

launch a web app from refUI

## Checklist

### DIFF
--- a/core/launcher/src/launcher_event_processor.rs
+++ b/core/launcher/src/launcher_event_processor.rs
@@ -25,20 +25,15 @@ use ripple_sdk::{
     async_trait::async_trait,
     extn::{
         client::{
-            self,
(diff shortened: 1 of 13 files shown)


satlead:
This PR was closed because Ripple cannot launch and manage focus between Web Firebolt vs Non Firebolt Web Apps and Native Apps  which would be needed for a System App like Accelerator UI to fully offload the App Launch operations. Our expectation in Ripple Launcher extension is to support launching Firebolt Web Apps like Accelerator UI and Certification App to provide Firebolt API support. In future if there is a launcher component and a focus manager component available as system components then Ripple would integrate with them and offer this services as part of the Firebolt Support for Launch operations.

PR #381, opened 2024-01-17
feat: PlatformBuildInfo contract

kpears201:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


device/thunder_ripple_sdk/src/lib.rs
```diff
@@ -67,4 +67,6 @@ pub mod tests {
         pub mod thunder_wifi_pacts;
         pub mod thunder_window_manager_pacts;
     }
+    pub mod mock_extension_client;
```
satlead, line 70: Can we build flag #[cfg(test)] this for a test utility?

device/thunder_ripple_sdk/src/tests/mock_extension_client.rs
```diff
@@ -0,0 +1,40 @@
...
+        extn_id::{ExtnClassId, ExtnId},
+        ffi::ffi_message::CExtnMessage,
+    },
+    framework::ripple_contract::RippleContract,
+    uuid::Uuid,
+};
+
+pub struct MockExtnClient {}
+
+impl MockExtnClient {
```
satlead, line 15: Can we move this to tdk. Actually @bsenth200  is doing this on another effort too. if we can move this to tdk it can be used by both

satlead (approved):
Some suggestions to port the code into separate layers

device/thunder_ripple_sdk/src/lib.rs
kpears201: it is, it is on the mod that this is under

device/thunder_ripple_sdk/src/tests/mock_extension_client.rs
kpears201: is there an existing place already?

pahearn73 (approved):

PR #382, opened 2024-01-17
fix: stop updating token in a loop

satlead:
## What

Changes for #379  Caused an infinite loop of token updates.

## Why

There was no check for tokens before updating it first time.

## How

Checking for the presence of tokens before updating.

## Test

## Checklist

### DIFF
--- a/core/main/src/processor/main_context_processor.rs
+++ b/core/main/src/processor/main_context_processor.rs
@@ -129,7 +129,7 @@ impl MainContextProcessor {
             });
         });
     }
-    pub async fn initialize_token(state: &PlatformState) {
+    pub async fn initialize_session(state: &PlatformState) {
         if !Self::check_account_session_token(state).await {
             error!("Account session still not available");
         } else if state.supports_cloud_sync() {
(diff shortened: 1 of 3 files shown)


core/main/src/processor/main_context_processor.rs
```diff
@@ -233,9 +214,18 @@ impl ExtnEventProcessor for MainContextProcessor {
         if let Some(update) = &extracted_message.update_type {
             match update {
                 RippleContextUpdateType::TokenChanged => {
-                    if let ActivationStatus::AccountToken(_t) = &extracted_message.activation_status
+                    if let ActivationStatus::AccountToken(t) = &extracted_message.activation_status
                     {
-                        Self::initialize_token(&state.state).await
+                        //Call initialize token only when account session was not initialized the first time
+                        if state.state.session_state.get_account_session().is_none() {
+                            Self::initialize_token(&state.state).await
```
pahearn73, line 221: Aren't we still calling `initialize_token` on startup, so this will potentially get called twice? There's stuff in there that is truly initialization i.e. do once.

kpears201 (approved):

core/main/src/processor/main_context_processor.rs
pahearn73: Should `initialize_token` be refactored to extract the `check_account_session_token` call?

satlead: yes we addressed this in the PR. 

1. During bootstrap if account_session is available we will just call ripple context update
2. During Token update we will first check if the account_session is already available if so we will just update token.
If account token is not available we will call initialize_token to make sure we kick start the process for token interactions

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #387, opened 2024-01-22
GetAdRouter: Changed to call GetXifa/GetAdRouter instead of GetAdInitObject

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


pahearn73:
I haven't removed support for AdvertisingRequest::GetAdInitObject calls because I wasn't sure if there might be plans to use it again in any way, any thoughts on this? If it's really not going to be used we can just get rid of it.

core/main/src/firebolt/handlers/advertising_rpc.rs
```diff
@@ -330,76 +338,107 @@ impl AdvertisingServer for AdvertisingImpl {
...
+
+        let advertising_request = AdvertisingRequest::GetAdConfig(AdConfigRequestParams {
+            privacy_data: privacy_data.clone(),
+            durable_app_id: durable_app_id.clone(),
             dist_session: session
+                .clone()
                 .ok_or_else(|| Error::Custom(String::from("no session available")))?,
-            scope: HashMap::new(),
+            environment: environment.to_string(),
+            scope: get_scope_option_map(&None),
```
pahearn73, line 367: Currently the firebolt `advertising.config` API doesn't include a `scope` parameter like the `advertising.id` API does, it seems like the spec should include it since `advertising.config` now gets the xifa directly. Was this a miss when `advertising.config` called `GetAdInitObject`?

kpears201: advertising.config will only return the default root scope, we don't support using advertising.config to get other scoped IFAs

kpears201 (approved):

Nv-Delle1416 (approved):

PR #406, opened 2024-01-25
App Management: Fixed race condition for ripple catalog update listener registration.

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/processor/main_context_processor.rs
+++ b/core/main/src/processor/main_context_processor.rs
@@ -176,16 +176,8 @@ impl MainContextProcessor {
 
             if state.supports_app_catalog() {
                 let ignore_list = vec![state.get_device_manifest().applications.defaults.main];
-
                 let client = state.get_client().clone();
                 client.add_event_processor(AppsUpdater::new(client.get_extn_client(), ignore_list));
-
-                let on_apps_update_resp = state
-                    .ripple_client
-                    .send_extn_request(AppCatalogRequest::OnAppsUpdate)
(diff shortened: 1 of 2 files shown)


core/main/src/processor/main_context_processor.rs
```diff
@@ -176,16 +175,8 @@ impl MainContextProcessor {
 
             if state.supports_app_catalog() {
                 let ignore_list = vec![state.get_device_manifest().applications.defaults.main];
-
-                let client = state.get_client().clone();
+                let client = state.get_client();
                 client.add_event_processor(AppsUpdater::new(client.get_extn_client(), ignore_list));
-
```
pahearn73, line 182: Moved to the extension so that it is processed after the extension has initialized and is able to. Should be done by the extension anyway since it sets up an extension listener thread internally.

core/main/src/processor/main_context_processor.rs
```diff
@@ -176,16 +175,8 @@ impl MainContextProcessor {
 
             if state.supports_app_catalog() {
                 let ignore_list = vec![state.get_device_manifest().applications.defaults.main];
-
```
pahearn73, line 179: Got rid of `AppCatalogRequest::OnAppsUpdate` altogether, unnecessary request layer now that we've moved to OSS.

kpears201 (approved):

B-Larsen (approved):

PR #410, opened 2024-01-30
Rdk 6.1 fixes for FCA test failures.

kvfasil:
## What
few FCA test cases like wifi.scan(), Authentication.token Apis are failing.

## Why
wifi.scan issue:
rdk6 introduced firebolt-supported ResidentAppUI as the home UI,
hence the FCA app runs as a third-party app. In the app manifest, the ResidentAppUI has all the necessary access (referred to as refui in the app manifest).

Authentication.token Apis issue:
we have removed the support for token from the manifest because our device does not support it. However, there have been recent changes on the Ripple side, where they added platform.token and distributor.token hence we can mock the expected response as a reference.

## How

wifi.scan issue:
The FCA app is now launched with the app ID "comcast.test.firecert". provided full access(exclusory).

Authentication.token Apis issue:
Added require contracts and capability in the manifest. also added extn processor processing auth.tocken queries.

## Test

Build and run FCA test cases.
1) wifi.scan should return list of available Wi-Fi networks.
2) Auth.token Api's must pass.

## Checklist

### DIFF
--- a/distributor/general/src/distributor_general_ffi.rs
+++ b/distributor/general/src/distributor_general_ffi.rs
@@ -42,13 +42,15 @@ use ripple_sdk::{
 use crate::{
     general_advertising_processor::DistributorAdvertisingProcessor,
     general_discovery_processor::DistributorDiscoveryProcessor,
+    general_distributor_token_processor::DistributorTokenProcessor,
     general_media_events_processor::DistributorMediaEventProcessor,
     general_metrics_processor::DistributorMetricsProcessor,
+    general_paltform_token_processor::PlatformTokenProcessor,
     general_permission_processor::DistributorPermissionProcessor,
     general_privacy_processor::DistributorPrivacyProcessor,
     general_securestorage_processor::DistributorSecureStorageProcessor,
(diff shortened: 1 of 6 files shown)


satlead (approved):

satlead:
Contents of this pr pertain to 

1. Distributor General Extension: This extension is used only by Video accelerator device test suites so they mock cloud firebolt provider behaviors like tokens, metrics, permissions and other cloud and service related contracts
2. The manifests used in this PR is also only used by Video Accelerator devices for the bootstrap.

All tokens and sensitive data in the processors are derived from Firebolt Open rpc examples.

pahearn73 (approved):

satlead:
For reference the token used in this PR is same as
https://github.com/rdkcentral/firebolt-apis/blob/abecdf192a07b9ee7cbbe4941e22fc4823f3dd64/src/openrpc/authentication.json#L77

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #417, opened 2024-02-08
Perf internet status

Karthick-Somasundaresan:
## What

This PR adds a feature to refresh ripple context on request (Internet status to start with). Also, each of the processors can set a prerequisite whether it needs internet or provisioning or both before processing the incoming request. If any of the prerequisites are not satisfied then the processor will send an error response even without going through actual processing. 

## Why

This change helps in reducing processing and bail out in case of known errors at an early stage. 

## How

Each processor fills in their prerequisite while implementing the ExtnStreamProcessor using the get_prerequisites method. 
## Test

This has been tested using making the internet a prerequisite for a processor and then manually making the internet unavailable and observing the requests are errored out even without going through the actual processing. 
## Checklist


core/main/src/bootstrap/extn/load_session_step.rs
```diff
@@ -42,14 +42,6 @@ impl Bootstep<BootstrapState> for LoadDistributorValuesStep {
             return Ok(());
         }
         MainContextProcessor::initialize_session(&s.platform_state).await;
-        // Session available during startup make sure the context is updated with the token for first time
```
pahearn73, line 45: Is this being done elsewhere now ripple initialization? We have new changes for DistP initialization that depend on receipt of a session context update on start-up.

satlead: +1 we need this for cases where token is available during startup

core/sdk/src/api/context.rs
```diff
@@ -76,27 +78,71 @@ impl RippleContext {
         RippleContext::get_from_payload(msg.clone())
     }
 
-    pub fn update(&mut self, request: RippleContextUpdateRequest) {
+    pub fn update(&mut self, request: RippleContextUpdateRequest) -> bool {
         match request {
             RippleContextUpdateRequest::Activation(a) => {
-                self.activation_status = a.into();
-                self.update_type = Some(RippleContextUpdateType::ActivationStatusChanged);
+                let activation_status: ActivationStatus = a.into();
```
pahearn73, line 84: Nitpick but you could probably simplify these, e.g. something like:

`
        let activation_status: ActivationStatus = a.into();
        if let Some(status) = self.activation_status {
            if status == &activation_status {
                return false
            }
        }
        self.activation_status = Some(activation_status);
        self.update_type = Some(RippleContextUpdateType::ActivationStatusChanged);
        true
`

...maybe similar for some of the other below as well.

core/main/src/bootstrap/extn/load_session_step.rs
Karthick-Somasundaresan: Yes, this state will be updated [here](https://github.com/rdkcentral/Ripple/blob/c3e23fd9668722fd3a7337e12d49595389151fc4/core/main/src/processor/main_context_processor.rs#L93).

core/main/src/processor/main_context_processor.rs
```diff
@@ -242,26 +293,32 @@ impl ExtnEventProcessor for MainContextProcessor {
...
-                            state
-                                .state
-                                .session_state
-                                .insert_session_token(t.token.clone())
-                        }
+                        state
+                            .state
+                            .session_state
+                            .insert_session_token(t.token.clone());
+                        Self::initialize_session(&state.state).await
```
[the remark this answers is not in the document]
pahearn73: I think as long as `Self::initialize_session` can never be called more than once, we're good -- thanks for the explanation 👍

Nv-Delle1416 (approved):

PR #422, opened 2024-02-12
feat: Support disabling uninstalls, support installing in priority order

kpears201:
## What

This adds support to be able to disable catalog uninstalls both from device_manifest and a remote feature control extension.
This adds support to first sort the installed apps by priority before installing them.
This adds a number of unit tests around apps updater.
This adds better support in MockExtnClient to be able mock main being able to make a call to a mock extn. In order to be able to use MockExtnClient in both main and sdk, it needs to be moved to the sdk under a tdk feature flag.

## Why

To support the roll out of the catalog management feature

## Test

## Checklist


core/main/src/service/apps/apps_updater.rs
```diff
@@ -126,18 +150,304 @@ pub async fn update(mut client: ExtnClient, apps_update: AppsUpdate, ignore_list
...
-        if !installed_ids.into_iter().any(|installed_app| {
-            installed_app.id.eq(&app.id) && installed_app.version.eq(&app.version)
-        }) {
-            debug!("update: Application is not currently installed, installing: title={}, id={}, version={}", app.title, app.id, app.version);
-            let metadata = AppMetadata::new(app.id, app.title, app.version, app.uri, app.data);
-            let resp = client.request(AppsRequest::InstallApp(metadata)).await;
-
-            if let Err(e) = resp {
-                error!("update: Could not install app: e={:?}", e);
+    let mut app_list: Vec<app_catalog::AppMetadata> = apps_update
```
kpears201, line 153: Refactored this using `filter` collection operation. This is so we can filter out the already installed apps before sorting.

core/main/src/service/apps/apps_updater.rs
```diff
@@ -92,28 +114,30 @@ pub async fn update(mut client: ExtnClient, apps_update: AppsUpdate, ignore_list
         }
     };
 
-    // Uninstall removed apps
+    if uninstalls_enabled {
```
kpears201, line 117: This is all whitespace changes, this just put all the uninstall code in a block that checks if it is enabled first.

core/main/src/processor/main_context_processor.rs
```diff
@@ -199,7 +199,20 @@ impl MainContextProcessor {
             if state.supports_app_catalog() {
                 let ignore_list = vec![state.get_device_manifest().applications.defaults.main];
                 let client = state.get_client();
-                client.add_event_processor(AppsUpdater::new(client.get_extn_client(), ignore_list));
+                let uninstalls_enabled = RemoteFeature::flag(
```
kpears201, line 202: Uses the new RemoteFeature impl code to fetch a uninstalled_enabled flag on boot and then pass that to the AppsUpdater module.

satlead: This manifest design patten is very scalable. We can probably work with Accelerator UI folks to define some for the VA Manifests. cc @arun-madhavan-013 @joseinweb

satlead (approved):

pahearn73 (approved):

PR #449, opened 2024-02-15
update coverage threshold script

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -50,11 +50,8 @@ jobs:
     steps:
       - name: read threshold file into environment variable
         run: |
-          git init
-          git clone https://github.com/rdkcentral/Ripple.git
-          cd Ripple
-          git fetch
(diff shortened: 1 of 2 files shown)


.github/workflows/Build.yml
```diff
@@ -48,12 +48,17 @@ jobs:
     name: Generate Code Coverage
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
       - name: read threshold file into environment variable
         run: |
+          git init
+          git clone https://github.com/rdkcentral/Ripple.git
```
satlead, line 54: In git there are options to juust get that branch alone as a single clone instead of the full set

git clone <url> --branch <branch> --single-branch <folder>

we can avoid 4 steps with 1

brendanobra (approved):

PR #460, opened 2024-02-29
add optional permission payload to permission request

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


satlead (approved):

SKumarMetro (commented):
Can't use existing enum CapabilityRole instead of creating another enum Role

maggie98choy:
Yes, we can use the existing enum CapabilityRole

Nv-Delle1416 (approved):

PR #463, opened 2024-03-01
Firebolt Permissions: Update app when not available

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/service/apps/pending_session_event_processor.rs
```diff
@@ -0,0 +1,81 @@
...
+            let has_pending_session = state.session_state.has_pending_session(&operation.id);
+
+            if has_pending_session {
+                if operation.success {
+                    // Send request to create session and send lifecyclemanagement.OnSessionTransitionComplete
+                    match PermissionHandler::fetch_permission_for_app_session(&state, &operation.id)
+                        .await
+                    {
+                        Ok(()) => {
+                            DelegatedLauncherHandler::emit_completed(&state, &operation.id).await
```
kpears201, line 65: doesn't there need to be more work for delegated launcher handler to do after it fetches the permissions? For example, it needs to check the grant states again. Basically everything that happens after the return statement you added needs to start executing again

kpears201 (approved):

B-Larsen (approved):

PR #469, opened 2024-03-06
Rust upgrade: 1.76.0

brendanobra:
Trying rust upgrade

## What

## Rust, brand new

Why are these changes needed?

## 1.69.0 is pretty old

## Upgrade

## Checklist

### DIFF
--- a/core/main/src/service/apps/delegated_launcher_handler.rs
+++ b/core/main/src/service/apps/delegated_launcher_handler.rs
@@ -146,14 +146,7 @@ impl AppManagerState {
             HashMap::new()
         }
     }
-    /*
-    Adding to suppress:
-    ```note: joining a path starting with separator will replace the path instead
-    help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#join_absolute_paths
-    note: `-D clippy::join-absolute-paths` implied by `-D warnings```
-    to avoid breaking any existing functionality
-    */
-    #[allow(clippy::join_absolute_paths)]
+
     fn get_storage_path(saved_dir: &str) -> String {
         let mut path = std::path::Path::new(saved_dir).join("app_info");
         if !path.exists() {
@@ -164,7 +157,7 @@ impl AppManagerState {
                     err
                 );
                 path =
-                    std::path::Path::new(&env::temp_dir().display().to_string()).join("/app_info");
+                    std::path::Path::new(&env::temp_dir().display().to_string()).join("app_info");
                 if let Err(err) = fs::create_dir_all(path.clone()) {
                     error!(
                         "Could not create directory {} for persisting app info err: {:?}, app title will persist in /tmp/",
--- a/core/main/src/service/apps/delegated_launcher_handler.rs
+++ b/core/main/src/service/apps/delegated_launcher_handler.rs
@@ -146,6 +146,14 @@ impl AppManagerState {
             HashMap::new()
         }
     }
+    /* 
+    Adding to suppress:
+    ```note: joining a path starting with separator will replace the path instead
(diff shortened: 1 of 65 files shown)


core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -146,6 +146,14 @@ impl AppManagerState {
...
         }
     }
+    /*
+    Adding to suppress:
+    ```note: joining a path starting with separator will replace the path instead
+    help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#join_absolute_paths
+    note: `-D clippy::join-absolute-paths` implied by `-D warnings```
+    to avoid breaking any existing functionality
+    */
+    #[allow(clippy::join_absolute_paths)]
```
pahearn73, line 156: Don't we just want to remove the `/` below in:
`std::path::Path::new(&env::temp_dir().display().to_string()).join("/app_info")`?

core/main/src/service/user_grants.rs
```diff
@@ -620,6 +623,7 @@ impl GrantState {
             "Incoming grant check for app_id: {:?} and role: {:?}",
             app_id, role_info
         );
+        #[allow(clippy::unnecessary_fallible_conversions)]
```
pahearn73, line 626: Can we use `from` instead of `try_from`?

core/main/src/service/apps/delegated_launcher_handler.rs
brendanobra: Maybe? I assume all these things are they way the are for a good reason, so did not want to arbitrarily change it.

core/main/src/service/user_grants.rs
brendanobra: i could not get it to compile, but probably a skillz issue on my side, and i was also a little concerned about why it was try_from the 1st place- - the conversion is infallible from the looks of it, but did not want to assume and break stuff.

core/main/src/service/apps/delegated_launcher_handler.rs
pahearn73: @Karthick-Somasundaresan : Was the `/` intentional for some reason?

core/main/src/service/user_grants.rs
pahearn73: I think this compiles. Maybe at one point role_info was a generic type or something hence the `try_from`?

`let permission = FireboltPermission::from(role_info);`

core/main/src/service/apps/delegated_launcher_handler.rs
brendanobra: No clue. The only reason this is popping up now is because clippy has become smarter between 1.69.0 and 1.76.0 . I did contemplate adding a unit test(s) , but that seems out of scope for this exercise

brendanobra: so.. is this actionable? fix it? leave it? would like to unblock the compiler upgrade if possible

pahearn73: I'm pretty sure line 167 is a bug because it will create /app_info and not <temp_dir>/app_info, I vote to just remove the `/` and remove `#[allow(clippy::join_absolute_paths)]`.

pahearn73 (approved):

core/main/src/service/apps/delegated_launcher_handler.rs
brendanobra: Thanks @pahearn73 ! Done

Nv-Delle1416 (approved):

PR #470, opened 2024-03-07
fix: Privacy data Call Optimization, caching privacy settings values

Vinodsathyaseelan:
## What

Privacy data Call Optimization, caching privacy settings values

## Why

In order to increase the resiliency and improve the latency of critical calls related to app launching we would require Ripple to Synchronize the values for privacy on boot and during linchpin updates in Platform State.

We expect some performance improvements in terms of latency for Privacy related api calls like
```
          Parameters.initialization
          Privacy Settings
          Badger calls related to privacy.
```

## How
Use cache for privacy settings storage properties 

## Test

## Checklist

### DIFF
--- a/core/main/src/processor/storage/storage_manager.rs
+++ b/core/main/src/processor/storage/storage_manager.rs
@@ -76,29 +76,20 @@ pub struct StorageManager;
 
 impl StorageManager {
     pub async fn get_bool(state: &PlatformState, property: StorageProperty) -> RpcResult<bool> {
-        if property.is_a_privacy_setting_property() {
-            // check if the privacy setting property is available in cache
-            let privacy_settings_cache = state.ripple_cache.get_privacy_settings_cache();
-            let val_opt = property.get_privacy_setting_value(&privacy_settings_cache);
-            if let Some(val) = val_opt {
-                return Ok(val);
-            }
+        if let Some(val) = state
+            .ripple_cache
+            .get_cached_bool_storage_property(&property)
+        {
+            return Ok(val);
         }
-
         let data = property.as_data();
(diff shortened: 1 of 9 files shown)


satlead (approved):
Looks good are there other items we want to add to the cache in future?

core/main/src/processor/storage/storage_manager.rs
```diff
@@ -76,11 +76,32 @@ pub struct StorageManager;
...
     pub async fn get_bool(state: &PlatformState, property: StorageProperty) -> RpcResult<bool> {
+        if property.is_a_privacy_setting_property() {
+            // check if the privacy setting property is available in cache
+            let privacy_settings_cache = state.ripple_cache.get_privacy_settings_cache();
+            let val_opt = property.get_privacy_setting_value(&privacy_settings_cache);
+            if let Some(val) = val_opt {
+                return Ok(val);
+            }
+        }
+
```
pahearn73, line 87: This should probably be made generic so privacy property-specific conditionals don't have to be inserted throughout `StorageManager`, the added bonus is that any arbitrary property would have cache support if we want. E.g. instead of `RippleCache::get_privacy_settings_cache()` you could have `RippleCache::get_cache(property)` and let `RippleCache` figure out if the particular property is cached.

Vinodsathyaseelan: @pahearn73 : Thanks for the review comment. 
RippleCache now handles the identification of cache-able properties and performs set/get values to/from the cache.

Vinodsathyaseelan:
> Looks good are there other items we want to add to the cache in future?

Yes. Added comments in RippleCache related to this.

pahearn73 (approved):

PR #472, opened 2024-03-07
Detect sideloading

kpears201:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/service/apps/apps_updater.rs
```diff
@@ -39,16 +144,46 @@ impl AppsUpdater {
...
             },
             streamer: DefaultExtnStreamer::new(),
         }
     }
+
+    pub async fn init(state: &PlatformState) -> AppsUpdater {
+        let ignore_list = vec![state.get_device_manifest().applications.defaults.main];
+        let mut extn_client = state.get_client().get_extn_client();
+        let feats = state.get_device_manifest().get_features();
+        let sideload_support = DeviceInfo::is_debug(&mut extn_client).await;
```
kpears201, line 169: Most of this method was moved out of main_context_processor to here.
I have added a check to query the device is a debug device before allowing the sideload logic.

core/sdk/src/api/manifest/device_manifest.rs
```diff
@@ -50,32 +50,28 @@ pub struct RippleConfiguration {
     pub distribution_platform: String,
     pub distribution_id_salt: Option<IdSalt>,
     pub form_factor: String,
-    #[serde(default = "default_values_default")]
+    #[serde(default)]
```
kpears201, line 53: I switched DeviceManifest and all its sub-structs to be `Default` capable. So that a unit test can just create a default manifest.
Instead of just doing defaults in serde, now serde just does the default in code.

core/sdk/src/api/manifest/device_manifest.rs
```diff
@@ -346,38 +342,32 @@ pub enum AppLauncherMode {
...
-        allow_product_analytics: false,
-        allow_remote_diagnostics: false,
-        allow_resume_points: false,
-        allow_unentitled_personalization: false,
-        allow_unentitled_resume_points: false,
-        allow_watch_history: false,
-        skip_restriction: "none".to_string(),
-        video_dimensions: default_video_dimensions(),
+impl Default for DefaultValues {
+    fn default() -> Self {
```
kpears201, line 346: This is all the same values, just put in a Default impl so it can be called from unit test instead of only when the manifest file is read.

core/main/src/service/apps/apps_updater.rs
```diff
@@ -11,22 +15,127 @@ use ripple_sdk::{
...
 
+use crate::state::platform_state::PlatformState;
+
 #[derive(Clone)]
 pub struct AppsUpdaterState {
     client: ExtnClient,
     ignore_list: Vec<String>,
     uninstalls_enabled: bool,
+    sideload_support: bool,
+    failed_app_installs: Arc<RwLock<Vec<FailedAppInstall>>>,
```
pahearn73, line 43: Any reason this is an array instead of a map with the app_id as the key? Then you wouldn't have to do all that iterating later.

core/main/src/service/apps/apps_updater.rs
```diff
@@ -72,31 +211,67 @@ impl ExtnStreamProcessor for AppsUpdater {
...
+fn is_expected_install(
+    old_catalog: &Option<Vec<app_catalog::AppMetadata>>,
+    failed_installs: &Arc<RwLock<Vec<FailedAppInstall>>>,
+    installed_app: &InstalledApp,
+) -> Option<bool> {
+    let in_old_catalog = old_catalog.as_ref().map(|apps| {
+        apps.iter()
+            .any(|a| a.id == installed_app.id && a.version == installed_app.version)
+    });
+    if in_old_catalog.is_some() && in_old_catalog.unwrap() {
```
pahearn73, line 243: Shouldn't this just be `if in_old_catalog.is_some() {`?

core/main/src/service/apps/apps_updater.rs
kpears201: yeah, it could be. I tend to just lean towards arrays instead of maps so it doesn't lock you into a primary key in that map. Like if we wanted to store multiple failed installed versions for the same app, then we would have to redo the keys. Probably not a big deal, but also think the iter methods are pretty easy to do a lookup by any field

core/main/src/service/apps/apps_updater.rs
```diff
@@ -72,31 +211,67 @@ impl ExtnStreamProcessor for AppsUpdater {
...
+/// Checks if the given installed app is
+/// the expected app to be installed
+/// If it is the same version in the old catalog
+/// or if it is the last_good_version before a failed install
+fn is_expected_install(
+    old_catalog: &Option<Vec<app_catalog::AppMetadata>>,
+    failed_installs: &Arc<RwLock<Vec<FailedAppInstall>>>,
+    installed_app: &InstalledApp,
+) -> Option<bool> {
+    let in_old_catalog = old_catalog.as_ref().map(|apps| {
```
pahearn73, line 239: I'm a little confused what the meaning of the `Option<bool>` return value is, can you explain or map it out in the function comment?

kpears201: We return None for when old_catalog is None. It basically means we don't know if this app is expected to be installed.
I guess technically based on the method name we should return false. If there is no old catalog, then NO apps should be installed. So the fact any app is installed means it is "not expected", but that breaks down when we go from softcat to distp. If this is the very first distp catalog we have installed, then the softcat version IS expected.
To make it simpler this method can return:
Yes it is expected, No it is not expected, I don't know if this is expected.
In the I don't know case we always choose to install the app from distp.

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #476, opened 2024-03-18
RPPL-1808:Read /etc/rippleversion.txt and update version

pkumbh631:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


patch
satlead, line 1: delete this patch

kpears201 (approved):

core/main/src/state/bootstrap_state.rs
```diff
@@ -107,9 +109,42 @@ impl BootstrapState {
...
-
+        let platform_state = PlatformState::new(
+            extn_manifest,
+            device_manifest,
+            client,
+            app_manifest_result,
+            ripple_version_from_etc(),
+        );
+
+        fn ripple_version_from_etc() -> Option<String> {
```
Vinodsathyaseelan, line 120: You may change this inner function as a closure 
`let ripple_version_from_etc = || -> Option<String>  {`

B-Larsen (approved):

PR #478, opened 2024-03-22
Features: Set defaults from manifest on boot

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/firebolt_ws.rs
+++ b/core/main/src/firebolt/firebolt_ws.rs
@@ -220,13 +220,12 @@ impl FireboltWs {
             error!("Error registering the connection {:?}", e);
             return;
         }
-        if !gateway_secure {
-            if PermissionHandler::fetch_and_store(&state, &app_id, false)
+        if !gateway_secure
+            && PermissionHandler::fetch_and_store(&state, &app_id, false)
                 .await
                 .is_err()
-            {
-                error!("Couldnt pre cache permissions");
-            }
+        {
+            error!("Couldnt pre cache permissions");
         }
 
         let (mut sender, mut receiver) = ws_stream.split();
--- a/core/main/src/firebolt/firebolt_ws.rs
+++ b/core/main/src/firebolt/firebolt_ws.rs
@@ -221,9 +221,6 @@ impl FireboltWs {
             return;
         }
         if !gateway_secure {
-            state
-                .session_state
-                .add_pending_session(app_id.clone(), None);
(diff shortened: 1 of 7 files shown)


core/main/src/firebolt/firebolt_ws.rs
```diff
@@ -221,9 +221,6 @@ impl FireboltWs {
             return;
         }
         if !gateway_secure {
-            state
-                .session_state
-                .add_pending_session(app_id.clone(), None);
```
pahearn73, line 226: Unrelated but pending sessions are added when perms are checked upon app launch, not needed here.

core/main/src/service/context_manager.rs
```diff
@@ -93,6 +94,14 @@ impl ContextManager {
...
         // Asynchronously get context and update the state
         tokio::spawn(async move {
+            // Set default cloud permissions value
+            ps_c.get_client().get_extn_client().context_update(
+                RippleContextUpdateRequest::UpdateFeatures(vec![FeatureUpdate::new(
+                    FEATURE_CLOUD_PERMISSIONS.into(),
+                    ps_c.get_device_manifest().get_features().cloud_permissions,
+                )]),
+            );
+
```
pahearn73, line 104: Set default cloud_permissions feature when ContextManager initially starts up, may eventually be overridden by RFC flag.

core/main/src/state/cap/permitted_state.rs
```diff
@@ -150,11 +150,15 @@ impl PermissionHandler {
         app_id: &str,
         allow_cached: bool,
     ) -> RippleResponse {
+        if state.open_rpc_state.is_app_excluded(app_id) {
+            return Ok(());
+        }
+
```
pahearn73, line 156: I actually recently added this for just `device_fetch_and_store`, but moved it here so it works for both cloud and device perms.

core/sdk/src/api/context.rs
```diff
@@ -141,14 +141,25 @@ impl RippleContext {
                 false
                 // This is not an update request so need not to honour it
             }
-            RippleContextUpdateRequest::Features(features) => {
+            RippleContextUpdateRequest::UpdateFeatures(features) => {
```
pahearn73, line 144: Improved this a bit so we can pass in one or more features and indicate whether or not they're en/disabled.

satlead (approved):

kpears201 (approved):

PR #487, opened 2024-03-27
Thunder plugin handling

satlead:
## What

The current handling of Thunder plugins does not take into account activation states

## Why

Thunder client cannot make calls to a plugin that is not yet activated.

## How

Established a new struct we can build this more going forward for any incoming messages

```#[derive(Debug, Deserialize)]
pub struct ThunderError {
    pub code: i32,
    pub message: String,
}

impl ThunderError {
    pub fn get_plugin_state(&self) -> PluginState {
        match self.message.as_str() {
            "ERROR_INPROGRESS" | "ERROR_PENDING_CONDITIONS" => PluginState::InProgress,
            "ERROR_UNKNOWN_KEY" => PluginState::Missing,
            _ => PluginState::Unknown,
        }
    }
}
```
2. Updated the Ready method to serve more value. This will now add timeout logic
```
pub async fn ready(self) { -> pub async fn ready(self) -> bool
```
3. Activate Plugin now returns the status for In progress
4. If the plugin is in error state i added a custom error message for the
```
json!({"error": "pre send error"}
```

5. Fixed the bad parsing inside thunder_device_info

## Test

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
+++ b/device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
@@ -15,9 +15,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use std::time::Duration;
+
 use ripple_sdk::{
     api::status_update::ExtnStatus,
-    log::{info, warn},
+    log::{error, info, warn},
     utils::error::RippleError,
 };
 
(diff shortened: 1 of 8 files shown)


device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
```diff
@@ -675,18 +679,28 @@ impl ThunderDeviceInfoRequestProcessor {
...
-            .parse()
-            .unwrap();
         let mut hdcp_response = HashMap::new();
-        if hdcp_version.contains("1.4") {
-            hdcp_response.insert(HdcpProfile::Hdcp1_4, is_hdcp_supported);
-        }
-        if hdcp_version.contains("2.2") {
-            hdcp_response.insert(HdcpProfile::Hdcp2_2, is_hdcp_supported);
+
+        if let Some(v) = response.message.get("supportedHDCPVersion") {
```
kpears201, line 684: For all of these can you just do an early return on None instead of putting the entire method inside a Some block?

device/thunder_ripple_sdk/src/client/plugin_manager.rs
```diff
@@ -272,39 +307,53 @@ impl PluginManager {
...
             .clone()
             .call(DeviceCallRequest {
                 method: Controller.method("activate"),
                 params: Some(DeviceChannelParams::Json(
                     serde_json::to_string(&r).unwrap(),
                 )),
             })
-            .await
+            .await;
+        if let Ok(plugin_error) = serde_json::from_value::<ThunderError>(resp.message) {
```
Vinodsathyaseelan, line 353: I couldn’t locate the code where we are responding DeviceCallRequest with ThunderError in resp.message. 
Are we setting `Value::Null` to resp.message on Thunder error ?

satlead: We create a custom internal error for the case where we didnt send the request to Thunder
```
json!({"error": "pre send error"}
```

device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
satlead: Followed this suggestion and now ThunderDevice Info has grown in to bit diff. Based on discussion with team will reduce the scope of thunder_device_info to a smaller fix for bootstrap related APIs

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
```diff
@@ -45,18 +47,48 @@ impl ThunderPoolStep {
...
         }
-        let controller_pool =
-            ThunderClientPool::start(url.clone(), None, thunder_connection_state.clone(), 1).await;
+        let controller_pool = ripple_sdk::tokio::time::timeout(
+            Duration::from_secs(10),
+            ThunderClientPool::start(url.clone(), None, thunder_connection_state.clone(), 1),
+        )
+        .await;
+
         if controller_pool.is_err() {
```
Vinodsathyaseelan, line 56: Do we need to the handle `Ok(Err(_)) `case here ?

kpears201: isn't that the next if block?
```
        let controller_pool = controller_pool.unwrap();
        if let Err(e) = controller_pool
```

device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
```diff
@@ -675,18 +663,32 @@ impl ThunderDeviceInfoRequestProcessor {
             })
             .await;
         info!("{}", response.message);
-        let hdcp_version = response.message["supportedHDCPVersion"].to_string();
-        let is_hdcp_supported: bool = response.message["isHDCPSupported"]
-            .to_string()
-            .parse()
-            .unwrap();
         let mut hdcp_response = HashMap::new();
+        let resp = response.message.get("supportedHDCPVersion");
```
kpears201, line 667: We really should just use serde at this point...

kpears201 (approved):

Nv-Delle1416 (approved):

PR #490, opened 2024-03-29
fix: Handling Thunder status for plugin

satlead:
## What

Increase the support for handling Thunder states provided by the status call for a Thunder plugin

## Why

Ripple has to be aware of the thunder state before making a call to Thunder.

## How

Including all the states listed under https://rdkcentral.github.io/Thunder/plugin/lifecycle/

## Test

Unit tests are included this will be tested in an environment where Thunder Plugins have a delayed startedup.

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/client/plugin_manager.rs
+++ b/device/thunder_ripple_sdk/src/client/plugin_manager.rs
@@ -83,10 +83,11 @@ impl ThunderError {
 impl PluginStatus {
     pub fn to_plugin_state(&self) -> PluginState {
         match self.state.as_str() {
-            "activated" => PluginState::Activated,
-            "resumed" => PluginState::Activated,
-            "suspended" => PluginState::Activated,
+            "activated" | "resumed" | "suspended" => PluginState::Activated,
             "deactivated" => PluginState::Deactivated,
+            "deactivation" => PluginState::Deactivation,
+            "activation" | "precondition" => PluginState::Activation,
+            "unavailable" | "hibernated" => PluginState::Unavailable,
             _ => PluginState::Missing,
         }
     }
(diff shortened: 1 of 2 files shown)


Nv-Delle1416 (approved):

device/thunder_ripple_sdk/src/client/plugin_manager.rs
```diff
@@ -83,10 +83,11 @@ impl ThunderError {
...
         match self.state.as_str() {
-            "activated" => PluginState::Activated,
-            "resumed" => PluginState::Activated,
-            "suspended" => PluginState::Activated,
+            "activated" | "resumed" | "suspended" => PluginState::Activated,
             "deactivated" => PluginState::Deactivated,
+            "deactivation" => PluginState::Deactivation,
+            "activation" | "precondition" => PluginState::Activation,
+            "unavailable" | "hibernated" => PluginState::Unavailable,
             _ => PluginState::Missing,
```
satlead, line 91: MIssing is probably not a true measure for a Plugin State as Thunder responds with Error message for callsigns which are not there in the system

satlead: Resolved it in the next commit

Vinodsathyaseelan (approved):

device/thunder_ripple_sdk/src/client/plugin_manager.rs
```diff
@@ -83,11 +83,11 @@ impl ThunderError {
 impl PluginStatus {
     pub fn to_plugin_state(&self) -> PluginState {
         match self.state.as_str() {
-            "activated" => PluginState::Activated,
-            "resumed" => PluginState::Activated,
-            "suspended" => PluginState::Activated,
+            "activated" | "resumed" | "suspended" => PluginState::Activated,
```
Vinodsathyaseelan, line 86: Need to check Thunder response for call request when the plugin is suspended state.

PR #491, opened 2024-03-29
fix: Add support for Dynamic Plugins

satlead:
## What

Some Thunder plugins are created dynamically. Ripple needs to understand these plugins and have support for their activation

## Why

To support Thunder plugins which are added dynamically to the framework.

## How

Uses the existing subscriber mechanism to setup state subscribers and provide the support.

## Test

Using a dynamic app plugin launched by RDKShell.launch

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/client/plugin_manager.rs
+++ b/device/thunder_ripple_sdk/src/client/plugin_manager.rs
@@ -134,6 +134,10 @@ pub enum PluginManagerCommand {
     ReactivatePluginState {
         tx: oneshot::Sender<PluginActivatedResult>,
     },
+    WaitForActivationForDynamicPlugin {
+        callsign: String,
+        tx: oneshot::Sender<PluginActivatedResult>,
+    },
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


device/thunder_ripple_sdk/src/client/thunder_client_pool.rs
```diff
@@ -234,6 +234,13 @@ mod tests {
                     PluginManagerCommand::WaitForActivation { callsign: _, tx } => {
                         oneshot_send_and_log(tx, PluginActivatedResult::Ready, "WaitForActivation");
                     }
+                    PluginManagerCommand::WaitForActivationForDynamicPlugin { callsign: _, tx } => {
```
Vinodsathyaseelan, line 237: Where is this command used ?

device/thunder_ripple_sdk/src/client/plugin_manager.rs
```diff
@@ -250,6 +254,10 @@ impl PluginManager {
                         let res = pm.reactivate_plugin_state().await;
                         oneshot_send_and_log(tx, res, "ReactivatePluginState");
                     }
+                    PluginManagerCommand::WaitForActivationForDynamicPlugin { callsign, tx } => {
```
Vinodsathyaseelan, line 257: Where is this command used ?

Vinodsathyaseelan (changes requested):
.

Vinodsathyaseelan (approved):

satlead:
Duplicate #503

PR #499, opened 2024-04-10
Discovery.launch api not returning errors when invalid NavigationIntent parameters passed for PlayEntity

Vinodsathyaseelan:
## What

Discovery.launch api not returning errors when invalid NavigationIntent parameters passed

## Why

Identified a few intent validation bugs related to PlayEntityIntent schema.

## How

Fixed PlayEntityIntent schema validation. 

## Test

Following unit test functions are used for schema validation

- test_playentity_intent_with_single_options
- test_playentity_intent_with_multiple_options
- test_playentity_intent_with_missing_options
- test_playentity_intent_with_invalid_options
- test_playentity_intent_with_options_and_program_entity_type
- test_playentity_intent_without_entity_type
- test_playentity_intent_movie_type_without_program_type
- test_playentity_intent_tv_episode_type_without_program_type
- test_playentity_intent_tv_episode_type_without_series_id
- test_playentity_intent_tv_episode_type_without_season_id
- test_playentity_intent_tv_episode_type_valid_case

## Checklist

### DIFF
--- a/core/sdk/src/api/device/entertainment_data.rs
+++ b/core/sdk/src/api/device/entertainment_data.rs
@@ -31,6 +31,7 @@ pub struct ContentProvider {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 #[serde(rename_all = "lowercase")]
 pub enum OfferingType {
     FREE,
@@ -51,6 +52,7 @@ impl Default for FederationOptions {
 }
 
 #[derive(Debug, Serialize, Deserialize, Default, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 #[serde(rename_all = "camelCase")]
 pub struct ContentIdentifiers {
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -66,6 +68,7 @@ pub struct ContentIdentifiers {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct ContentRating {
     pub scheme: SchemeValue,
     #[serde(deserialize_with = "rating_format_extender")]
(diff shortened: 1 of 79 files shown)


core/sdk/src/api/device/entertainment_data.rs
```diff
@@ -662,9 +662,48 @@ pub enum ProgramEntityIntentData {
     Extra(AdditionalEntity),
 }
 
+// PlayableProgramEntity is almost similar to ProgramEntityIntentData.
+// firebolt-open-rpc.json has used a subset of ProgramEntity as PlayableEntity
+// Need to check if we can merge them
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[serde(tag = "programType", rename_all = "camelCase")]
+pub enum PlayableProgramEntity {
```
kpears201, line 670: This is what I see in the docs:
```
type PlayableEntity =
  | MovieEntity
  | TVEpisodeEntity
  | PlaylistEntity
  | MusicEntity
  | AdditionalEntity
```

Shouldn't some of these be removed?

core/sdk/src/api/device/entertainment_data.rs
```diff
@@ -662,9 +662,48 @@ pub enum ProgramEntityIntentData {
...
+    pub entity_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub asset_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub app_content_data: Option<AppContentDataString>,
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct ProgramBaseEntity {
```
pahearn73, line 706: Just curious why you need to define both `BaseEntity` and `ProgramBaseEntity` since they seem to be the same thing (except that `entity_type` of the former is a `String` and the latter is a `ProgramEntityType`, but that's really just a string with a default value).

Vinodsathyaseelan: The default value for `entity_type` to "program" should be set only for a subset of entities which I categorized as `ProgramBaseEntity`.   The earlier implementation (BaseEntity) used to set the default entity_type value of all entities to `program`.  For example `playlist` is another valid entity_type.

kpears201 (approved):

Nv-Delle1416 (approved):

PR #501, opened 2024-04-11
feat: rdk telemetry logging

brendanobra:
## What

Injection of app version where available into telemetry 

## Why

to have app version in telemetry

## How

adding app version

## Test

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs
+++ b/device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs
@@ -537,6 +537,8 @@ impl ThunderPackageManagerRequestProcessor {
         let metrics_timer = start_service_metrics_timer(
             &thunder_state.get_client(),
             ThunderMetricsTimerName::PackageManagerGetList.to_string(),
+            None,
+            None,
         );
 
         let device_response = thunder_state
@@ -610,6 +612,8 @@ impl ThunderPackageManagerRequestProcessor {
         let metrics_timer = start_service_metrics_timer(
             &state.thunder_state.get_client(),
             ThunderMetricsTimerName::PackageManagerInstall.to_string(),
+            Some(app.clone().id),
+            Some(app.clone().version),
         );
 
         let device_response = state
@@ -642,7 +646,7 @@ impl ThunderPackageManagerRequestProcessor {
             Ok(handle) => {
                 let operation = Operation::new(
                     AppsOperationType::Install,
-                    app.id,
+                    app.clone().id,
                     AppData::new(app.version),
                 );
                 Self::add_or_remove_operation(state.clone(), handle.clone(), operation);
@@ -665,7 +669,8 @@ impl ThunderPackageManagerRequestProcessor {
         ) {
             info!(
                 "uninstall_app: Uninstallation already in progress: app={}, version={}",
-                app.id, app.version
+                app.clone().id,
+                app.clone().version
             );
 
             return Self::respond(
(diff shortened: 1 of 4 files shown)


satlead (approved):

device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs
```diff
@@ -640,12 +671,14 @@ impl ThunderPackageManagerRequestProcessor {
...
         match thunder_resp {
             Ok(handle) => {
-                let operation = Operation::new(
-                    AppsOperationType::Install,
-                    app.id,
-                    AppData::new(app.version),
-                );
+                let operation_timer = operation.timer.clone().stop();
                 Self::add_or_remove_operation(state.clone(), handle.clone(), operation);
+                stop_and_send_service_metrics_timer(
```
pahearn73, line 676: I don't think you want to stop the operation timer here, you want to stop it when we get a notification that an operation has completed (successfully or not). That happens [here](https://github.com/rdkcentral/Ripple/blob/main/device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs#L387), but the other situation you have to account for is that sometimes an operation will complete before we even get the `Ok(handle)` response from thunder in this block. Because of this I wonder if instead you want to conditionally start/stop the operation timer in [add_or_remove_operation()](https://github.com/rdkcentral/Ripple/blob/main/device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs#L464), which figures out whether to start the operation time-out timer depending on whether the operation is being added or removed (see the comment tombstone).

satlead:
Can we repurpose thunder_telemetry.rs to just print the log markers?

pahearn73 (approved):

pahearn73 (approved):

satlead (approved):

device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs
```diff
@@ -472,20 +491,28 @@ impl ThunderPackageManagerRequestProcessor {
...
+        /*
+        clone active_operations, we want an immutable view
+        */
+        let active_operations = state.active_operations.lock().unwrap().clone();
+        match active_operations.get(&handle) {
+            Some(op) => {
+                let mut timer = op.timer.clone();
+                timer.stop();
+                timer.insert_tag("status".to_string(), status.unwrap_or("".to_string()));
+                rdk_telemetry_emit(timer);
```
pahearn73, line 505: I think you need to remove the operation in this case.

device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs
```diff
@@ -472,20 +491,28 @@ impl ThunderPackageManagerRequestProcessor {
...
+            }
+            None => {
+                /*new operation, insert into map */
+                state
```
pahearn73, line 509: Instead of locking again here can you just use `active_operations` and don't clone it on line 499?

device/thunder_ripple_sdk/src/processors/thunder_package_manager.rs
```diff
@@ -510,20 +537,41 @@ impl ThunderPackageManagerRequestProcessor {
...
-                .is_some()
-            {
-                error!("Detected incomplete operation after {} seconds, attempting to cancel: handle={}", timeout_secs, handle.clone());
-                Self::cancel_operation(state, handle).await;
-            }
+            /*
+            clone active_operations, we want an immutable view, and stop/emit timer if the operation is in the map
+            */
+
+            let active_operations = state.active_operations.lock().unwrap().clone();
```
pahearn73, line 548: Same question re:cloning from [here](https://github.com/rdkcentral/Ripple/pull/501/files#r1581348471).

B-Larsen (approved):

PR #502, opened 2024-04-11
fix: disable package manager contract test - init

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


Vinodsathyaseelan (approved):

device/thunder_ripple_sdk/src/tests/contracts/thunder_package_manager_pacts.rs
```diff
@@ -366,169 +365,169 @@ async fn test_get_installed_apps() {
...
-
-    let resp = ThunderPackageManagerRequestProcessor::process_request(
-        tps,
-        msg,
-        AppsRequest::InstallApp(app_metadata_params),
-    )
-    .await;
-    assert!(resp);
-}
+// #[tokio::test(flavor = "multi_thread")]
```
satlead, line 368: Can we just not add an ignore instead of commenting?

satlead (approved):

PR #505, opened 2024-04-15
feat: Ripple Gateway Pass through

satlead:
## What

This change adds Communication Broker and a Rules engine to Ripple providing support for Ripple as an API Gateway with passthrough support.
https://wiki.rdkcentral.com/display/RDK/Ripple+as+a+API+Gateway

## Why

1. Provide faster rollout of new Firebolt API without a Ripple update.
2. Most of the existing Thunder APIs are Firebolt ready(FiRe) with some trivial JSON translation. The objective is to adopt a widely accepted JSON transformation language like jq, which allows operators to learn and configure rules without our support and training.
3. Provide a migration plan to include new New Thunder Plugins(FiRe) that become available in open source.
4. Should inherently support Firebolt Extension SDKs.

## How

By using Jq rules syntax to apply translations between Ripple and FiRe providers.

## Test

To test this feature a ripple.common.rules.json is added for some known Firebolt APIs which have a FiRe Thunder API companion. To test this add the file path on to the extn.json in the rules_path field. 

## Checklist


core/main/Cargo.toml
```diff
@@ -50,6 +50,12 @@ env-file-reader = "0.2.0"
 sd-notify = { version = "0.4.1", optional = true }
 exitcode = "1.1.2"
 rand = "0.8"
+url = "=2.5.0"
```
Vinodsathyaseelan, line 53: url version "2.2.2"  is defined in Ripple workspace. We may upgrade that and add workspace dependency here.

core/main/src/bootstrap/boot.rs
```diff
@@ -46,25 +47,29 @@ use super::{
...
 pub async fn boot(state: BootstrapState) -> RippleResponse {
     let bootstrap = Bootstrap::new(state);
+    execute_step(StartCommunicationBroker, &bootstrap).await?;
     execute_step(SetupExtnClientStep, &bootstrap).await?;
     execute_step(LoadExtensionMetadataStep, &bootstrap).await?;
     execute_step(LoadExtensionsStep, &bootstrap).await?;
     execute_step(StartExtnChannelsStep, &bootstrap).await?;
     execute_step(StartAppManagerStep, &bootstrap).await?;
     execute_step(LoadDistributorValuesStep, &bootstrap).await?;
+    execute_step(StartOtherBrokers, &bootstrap).await?;
```
Vinodsathyaseelan, line 72: What exactly is `OtherBroker` ? An example in the document section might  help.

core/main/Cargo.toml
```diff
@@ -50,6 +50,12 @@ env-file-reader = "0.2.0"
 sd-notify = { version = "0.4.1", optional = true }
 exitcode = "1.1.2"
 rand = "0.8"
+url.workspace = true
+futures-util = { version = "0.3.28", default-features = false, features = ["sink", "std"] }
```
pahearn73, line 54: I see `futures-util` imported but not actually used anywhere, am I missing something?

core/main/src/bootstrap/start_communication_broker.rs
```diff
@@ -0,0 +1,66 @@
...
+            BrokerOutputForwarder::start_forwarder(ps.clone(), rx)
+        }
+        // Setup the endpoints from the manifests
+        let mut endpoint_state = ps.endpoint_state;
+        endpoint_state.build_thunder_endpoint();
+        Ok(())
+    }
+}
+
+pub struct StartOtherBrokers;
```
pahearn73, line 46: What are the "other brokers" outside of the communication broker? Any reason why multiple "other" brokers are grouped together here instead of separated like the communication broker?

core/main/src/bootstrap/boot.rs
```diff
@@ -46,25 +47,29 @@ use super::{
...
-/// 1. [SetupExtnClientStep] - Initializes the extn client to start the Inter process communication backbone
-/// 2. [LoadExtensionMetadataStep] - Loads the Extn metadata from the So files
-/// 3. [LoadExtensionsStep] - Loads the Extensions in to [crate::state::extn_state::ExtnState]
-/// 4. [StartExtnChannelsStep] - Starts the Device channel extension
-/// 5. [StartAppManagerStep] - Starts the App Manager and other supporting services
-/// 6. [LoadDistributorValuesStep] - Loads the values from distributor like Session
-/// 7. [CheckLauncherStep] - Checks the presence of launcher extension and starts default app
-/// 8. [StartWsStep] - Starts the Websocket to accept external and internal connections
-/// 9. [FireboltGatewayStep] - Starts the firebolt gateway and blocks the thread to keep it alive till interruption.
+/// 1. [StartCommunicationBroker] - Initialize the communication broker to create Thunder broker if rules are setup.
```
pahearn73, line 50: Is there a sequence diagram somewhere that describes the sequence as a request passes through the communication broker, through the thunder broker, endpoint broker, etc.?

core/main/src/broker/endpoint_broker.rs
```diff
@@ -0,0 +1,598 @@
...
+/// back to the gateway for client consumption
+#[derive(Clone, Debug)]
+pub struct BrokerCallback {
+    pub sender: Sender<BrokerOutput>,
+}
+
+static ATOMIC_ID: AtomicU64 = AtomicU64::new(0);
+
+impl BrokerCallback {
+    /// Default method used for sending errors via the BrokerCallback
```
pahearn73, line 79: Is this is the "default" method used to send errors does that mean there's a way to override it with another method?

core/main/src/broker/endpoint_broker.rs
```diff
@@ -0,0 +1,598 @@
...
+    pub async fn send_error(&self, request: BrokerRequest, error: RippleError) {
+        let value = serde_json::to_value(JsonRpcError {
+            code: JSON_RPC_STANDARD_ERROR_INVALID_PARAMS,
+            message: format!("Error with {:?}", error),
+            data: None,
+        })
+        .unwrap();
```
pahearn73, line 86: Unsafe `unwrap`.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -0,0 +1,598 @@
...
+pub struct EndpointBrokerState {
+    endpoint_map: Arc<RwLock<HashMap<String, BrokerSender>>>,
+    callback: BrokerCallback,
+    request_map: Arc<RwLock<HashMap<u64, BrokerRequest>>>,
+    extension_request_map: Arc<RwLock<HashMap<u64, ExtnMessage>>>,
+    rule_engine: RuleEngine,
+}
+
+impl EndpointBrokerState {
+    pub fn get(tx: Sender<BrokerOutput>, rule_engine: RuleEngine) -> Self {
```
pahearn73, line 160: `get` implies return of a new or existing instance, but this is really a constructor so should be named `new` or similar.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -0,0 +1,598 @@
...
+        }
+    }
+
+    fn get_request(&self, id: u64) -> Result<BrokerRequest, RippleError> {
+        let result = { self.request_map.read().unwrap().get(&id).cloned() };
+        if result.is_none() {
+            return Err(RippleError::InvalidInput);
+        }
+
+        let result = result.unwrap();
```
pahearn73, line 176: Can be re-written to avoid unwrapping.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -0,0 +1,598 @@
...
+            let mut extn_map = self.extension_request_map.write().unwrap();
+            let _ = extn_map.insert(id, extn_message.unwrap());
+        }
+
+        rpc_request_c.ctx.call_id = id;
+        self.get_broker_request(&rpc_request_c, rule)
+    }
+
+    pub fn build_thunder_endpoint(&mut self) {
+        if let Some(endpoint) = self.rule_engine.rules.endpoints.get("thunder").cloned() {
```
pahearn73, line 221: Where are the keys like 'thunder" defined? Should they be listed explicitly via an enum or something?

Vinodsathyaseelan (approved):
The code in general looks good to me. However, I didn’t review any aspect of the component interactions and the run time business logic.

core/main/Cargo.toml
satlead: Thunder broker uses the sink and std capabilities for doing Websocket operations so it is internally used by the tokio tungstenite.

satlead: `use futures_util::{SinkExt, StreamExt};`

core/main/src/bootstrap/start_communication_broker.rs
satlead: @Vinodsathyaseelan had the same concern they are detailed in the document here https://wiki.rdkcentral.com/display/RDK/Ripple+as+a+API+Gateway#RippleasaAPIGateway-Brokers

core/main/src/bootstrap/boot.rs
satlead: Yes there is a sequence diagram i will add it to the Ripple as a Gateway documentation[ ](https://wiki.rdkcentral.com/display/RDK/Ripple+as+a+API+Gateway#RippleasaAPIGateway-Brokers)

core/main/src/broker/endpoint_broker.rs
```diff
@@ -0,0 +1,598 @@
...
+pub struct BrokerRequest {
+    pub rpc: RpcRequest,
+    pub rule: Rule,
+}
+
+/// BrokerCallback will be used by the communication broker to send the firebolt response
+/// back to the gateway for client consumption
+#[derive(Clone, Debug)]
+pub struct BrokerCallback {
+    pub sender: Sender<BrokerOutput>,
```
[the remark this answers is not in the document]
satlead: Yeah given we can have both BrokerResponse and BrokerEvent. BrokerOutput seemed more appropriate

core/main/src/broker/endpoint_broker.rs
satlead: This is an utility method would be useful as a last resort for
1. Invalid Rule
2. Invalid Endpoint
3. Rule compilation errors
4. Rule Filter errors
5. Timeout cases(future)

core/main/src/broker/endpoint_broker.rs
satlead: Values are predefined here 
```
let value = serde_json::to_value(JsonRpcError {
            code: JSON_RPC_STANDARD_ERROR_INVALID_PARAMS,
            message: format!("Error with {:?}", error),
            data: None,
        })
        .unwrap();
```

core/main/src/broker/endpoint_broker.rs
satlead: Yes, the new architecture expects Thunder to be default supported outside the SDK. Extensions for Thunder(and SDK) might not be needed in the future based on this new architecture

core/main/src/broker/endpoint_broker.rs
satlead: To avoid concurrent lock scenarios between the read and the write(cleanup for request/response cases). result is checked for `is_some` before the unwrap so it is still safe.

core/main/src/broker/rules_engine.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct RuleEngine {
+    pub rules: RuleSet,
+}
+
+impl RuleEngine {
+    fn build_path(path: &str, default_path: &str) -> String {
+        if path.starts_with('/') {
+            path.to_owned()
+        } else {
+            format!("{}{}", default_path, path)
```
pahearn73, line 113: If `default_path` doesn't end with `/` or path doesn't start with `/` this will not construct the path correctly. I think you can do:

`format!("{}{}", default_path, path)`

...since a path like `/some_directory//some_subdirectory` is still a valid path.

core/main/src/broker/endpoint_broker.rs
pahearn73: Yes but we've purposefully been removing this style from the codebase I think in part to satisfy clippy. There was a big effort to clean-up all repos.

core/main/src/broker/rules_engine.rs
```diff
@@ -0,0 +1,205 @@
...
+        return Err(RippleError::RuleError);
+    }
+
+    let inputs = RcIter::new(core::iter::empty());
+
+    // iterator over the output values
+    let mut out = f.run((Ctx::new([], &inputs), Val::from(input)));
+
+    if let Some(Ok(v)) = out.next() {
+        info!(
```
pahearn73, line 196: Should this be `debug!`?

core/main/src/broker/thunder_broker.rs
```diff
@@ -0,0 +1,222 @@
...
+            info!("Broker Endpoint url {}", endpoint.url);
+            let url = url::Url::parse(&endpoint.url).unwrap();
+            let port = extract_tcp_port(&endpoint.url);
+            info!("Url host str {}", url.host_str().unwrap());
+            //let tcp_url = url.host_str()
+            let tcp = loop {
+                if let Ok(v) = TcpStream::connect(&port).await {
+                    break v;
+                } else {
+                    error!("Broker Wait for a sec and retry {}", port);
```
pahearn73, line 73: I wonder if this shouldn't loop indefinitely or always log until we successfully connect, by logging every second we risk rotating and potentially losing logs if for some reason the device never makes a connection.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -0,0 +1,598 @@
...
+
+    /// Default handler method for the broker to remove the context and send it back to the
+    /// client for consumption
+    fn handle_response(result: &[u8], callback: BrokerCallback) {
+        let mut final_result = Err(RippleError::ParseError);
+        if let Ok(data) = serde_json::from_slice::<JsonRpcApiResponse>(result) {
+            final_result = Ok(BrokerOutput { data });
+        }
+        if let Ok(output) = final_result {
+            tokio::spawn(async move { callback.sender.send(output).await });
```
pahearn73, line 392: Does this need to spawn a thread?

core/main/src/broker/thunder_broker.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Default handler method for the broker to remove the context and send it back to the
+    /// client for consumption
+    fn handle_response(result: &[u8], callback: BrokerCallback) {
+        let mut final_result = Err(RippleError::ParseError);
+        if let Ok(data) = serde_json::from_slice::<JsonRpcApiResponse>(result) {
+            let updated_data = Self::update_response(&data);
+            final_result = Ok(BrokerOutput { data: updated_data });
+        }
+        if let Ok(output) = final_result {
+            tokio::spawn(async move { callback.sender.send(output).await });
```
pahearn73, line 217: Does this need to spawn a thread?

core/main/src/broker/thunder_broker.rs
```diff
@@ -0,0 +1,222 @@
...
+            let request = Self::update_request(rpc_request)?;
+            requests.push(request)
+        }
+
+        Ok(requests)
+    }
+
+    /// Default handler method for the broker to remove the context and send it back to the
+    /// client for consumption
+    fn handle_response(result: &[u8], callback: BrokerCallback) {
```
pahearn73, line 210: There's a duplicate of this in `endpoint_broker.rs`, consider making common.

core/main/src/broker/websocket_broker.rs
```diff
@@ -0,0 +1,110 @@
...
+    tokio::{self, net::TcpStream, sync::mpsc},
+};
+use std::time::Duration;
+use tokio_tungstenite::client_async;
+
+pub struct WebsocketBroker {
+    sender: BrokerSender,
+}
+
+fn extract_tcp_port(url: &str) -> String {
```
pahearn73, line 34: Duplicate in `thuner_broker.rs`.

core/main/src/broker/websocket_broker.rs
```diff
@@ -0,0 +1,110 @@
...
+        Self { sender: broker }
+    }
+}
+
+impl EndpointBroker for WebsocketBroker {
+    fn get_broker(endpoint: RuleEndpoint, callback: BrokerCallback) -> Self {
+        Self::start(endpoint, callback)
+    }
+
+    fn get_sender(&self) -> BrokerSender {
```
pahearn73, line 107: All implementers do the same thing, maybe just add the method to the trait.

core/main/Cargo.toml
pahearn73: Yes but I didn't see `SinkExt` or `StreamExt` used, maybe I missed that.

brendanobra (approved):

core/main/src/broker/rules_engine.rs
satlead: Think we avoided that strategy for cases with Windows development environment

core/main/src/broker/rules_engine.rs
satlead: used as RDK Telemetry marker which is needed for even Prod builds

core/main/src/broker/endpoint_broker.rs
satlead: Yes to release the queue for the next response

core/main/src/broker/thunder_broker.rs
satlead: Not exactly the same, Thunder broker does `let updated_data = Self::update_response(&data);` on top of the handle_response. Which is necessary to convert event messages  which have different JSON Rpc structure between Thunder and Firebolt

core/main/src/broker/websocket_broker.rs
satlead: Yes each implementer gets a unique broker sender which is used by endpoint broker state to send the request. 

BrokerSender is self sized to each instance of the Broker type. Super trait cannot hold instance of BrokerSender without having method definition for the same and it also needs a Self sized instance. This implementation is similar to the ExtnProcessor trait.

core/main/src/broker/endpoint_broker.rs
satlead: There are no clippy errors for this particular use case or the PR based on the GitHub Actions run. Having separate read and write blocks is necessary for the Lock safety.

core/main/src/broker/thunder_broker.rs
satlead: Updated the logic to print every 10secs with total secs taken for retry and the port

core/main/src/broker/websocket_broker.rs
satlead: Moved it to utils

core/main/src/broker/rules_engine.rs
pahearn73: Regardless of how you want to do it I think you should account for a missing `/`.

pahearn73 (approved):

PR #529, opened 2024-05-22
fix: advertising config

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/advertising_rpc.rs
+++ b/core/main/src/firebolt/handlers/advertising_rpc.rs
@@ -363,10 +363,6 @@ impl AdvertisingServer for AdvertisingImpl {
             .await
             .unwrap_or_default();
 
-        if country_code != "US" {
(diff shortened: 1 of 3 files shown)


core/main/src/firebolt/handlers/advertising_rpc.rs
```diff
@@ -359,6 +359,14 @@ impl AdvertisingServer for AdvertisingImpl {
         )
         .await;
 
+        let country_code = StorageManager::get_string(&self.state, StorageProperty::CountryCode)
```
Vinodsathyaseelan, line 362: Looks like we are repeating this code sequence/logic in multiple places.  You may move this to the code that provides `privacy_data` value.

kpears201 (approved):

Vinodsathyaseelan (approved):

B-Larsen (approved):

PR #530, opened 2024-05-23
set default value for audioDescriptionEnabled

maggie98choy:
## What

Set a default value for accessibility.audioDescriptionSettings firebolt API

## Why

Ripple StorageManager to return a default value for audioDescriptionEnabled when the corresponding entry is not available in Thunder Persistent Storage.

## How

## Test

## Checklist

### DIFF
--- a/core/sdk/src/api/manifest/device_manifest.rs
+++ b/core/sdk/src/api/manifest/device_manifest.rs
@@ -375,6 +375,8 @@ impl Default for DefaultValues {
             video_dimensions: default_video_dimensions(),
             media_progress_as_watched_events: false,
             accessibility_audio_description_settings: false,
+            accessibility_audio_description_settings:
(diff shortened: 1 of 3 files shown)


core/sdk/src/api/manifest/device_manifest.rs
```diff
@@ -372,6 +377,7 @@ impl Default for DefaultValues {
             allow_watch_history: false,
             skip_restriction: "none".to_string(),
             video_dimensions: default_video_dimensions(),
+            accessibility_audio_description_settings: false,
```
Vinodsathyaseelan, line 380: Use default_accessibility_audio_description_settings()  here, otherwise remove this function.

Vinodsathyaseelan (approved):

Nv-Delle1416 (approved):

PR #533, opened 2024-05-24
feat: Support for resume points using metrics.mediaprogress

Vinodsathyaseelan:
## What

Support for resume points using `metrics.mediaprogress` API.
This feature is turned off by default. 

Set the device manifest flag `mediaProgressAsWatchedEvents` to true under `default_values` section to turn on this feature.

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/metrics_rpc.rs
+++ b/core/main/src/firebolt/handlers/metrics_rpc.rs
@@ -25,8 +25,10 @@ use std::collections::HashMap;
 
 use ripple_sdk::{
     api::{
+        account_link::{AccountLinkRequest, WatchedRequest},
         firebolt::{
             fb_capabilities::JSON_RPC_STANDARD_ERROR_INVALID_PARAMS,
+            fb_discovery::WatchedInfo,
             fb_metrics::{
                 self, hashmap_to_param_vec, Action, BehavioralMetricContext,
                 BehavioralMetricPayload, CategoryType, ErrorParams, FlatMapValue,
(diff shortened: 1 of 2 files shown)


core/main/src/firebolt/handlers/metrics_rpc.rs
```diff
@@ -508,6 +510,35 @@ impl MetricsServer for MetricsImpl {
...
+            .media_progress_as_watched_events
+            && progress != MediaPositionType::None
+        {
+            let request = WatchedRequest {
+                context: ctx.clone(),
+                info: WatchedInfo {
+                    entity_id: media_progress_params.entity_id.clone(),
+                    progress: media_progress_params
+                        .progress
+                        .expect("progress shall not be None at this point"),
```
pahearn73, line 528: I feel like this should be written so `expect` isn't used here, it's similar to doing:

```
if some_option.is_none() {
  return;
}

do_something(some_option.unwrap());
```

...which we've since stopped doing in our code.

Also, could [watched()](https://github.com/rdkcentral/Ripple/blob/ad1f9c30581178635a820e4724a412a182dcaa59/core/main/src/firebolt/handlers/discovery_rpc.rs#L468) just be called instead of duplicating the code, we could also log an error if the request fails.

pahearn73: Understood but my point was `expect` is much like `unwrap` which we've been cleaning up.

I think we call other rpc handlers outside of the associated module for similar reasons of code duplication. Could you at least add a log if the call fails?

Vinodsathyaseelan: As suggested, removed the usage of `expect` . Added error log messages on failure to send watched event.

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #535, opened 2024-05-29
LifecycleManagement.setState - Added state transition validation 

Vinodsathyaseelan:
## What

Added app lifecycle state transition validation.

## Why

Current implementation allows state transition to `suspended` or `unloading` directly from `foreground`/`background`.
This has been fixed by including a state transition validation function

## How

Added the function `fn is_valid_lifecycle_transition(from: LifecycleState, to: LifecycleState) -> bool` to validate the app lifecycle state transition.

This validation check is disabled by default.  Set the feature flag `"lifecycle_transition_validate":true`  in device manifest to enable this check. 

## Test

## Checklist

### DIFF
--- a/core/main/src/service/apps/delegated_launcher_handler.rs
+++ b/core/main/src/service/apps/delegated_launcher_handler.rs
@@ -1310,17 +1310,192 @@ impl DelegatedLauncherHandler {
     }
 
     fn is_valid_lifecycle_transition(from: LifecycleState, to: LifecycleState) -> bool {
-        match to {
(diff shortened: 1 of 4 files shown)


core/main/src/service/apps/delegated_launcher_handler.rs
satlead, line 1: Very clean and awesome use of matching arms

satlead (approved):

B-Larsen (approved):

PR #536, opened 2024-05-30
Modify device uid support app migration

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/device_rpc.rs
+++ b/core/main/src/firebolt/handlers/device_rpc.rs
@@ -15,7 +15,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use std::{collections::HashMap, time::Duration};
+use std::{collections::HashMap, env, time::Duration};
 
 use crate::{
     firebolt::rpc::RippleRPCProvider,
@@ -28,6 +28,8 @@ use crate::{
 use jsonrpsee::{
     core::{async_trait, RpcResult},
     proc_macros::rpc,
+    tracing::debug,
+    types::error::CallError,
     RpcModule,
 };
 use ripple_sdk::{
@@ -47,6 +49,7 @@ use ripple_sdk::{
         },
         distributor::distributor_encoder::EncoderRequest,
         firebolt::{
+            fb_capabilities::CAPABILITY_NOT_SUPPORTED,
             fb_general::{ListenRequest, ListenerResponse},
             fb_openrpc::FireboltSemanticVersion,
         },
@@ -59,6 +62,7 @@ use ripple_sdk::{
     extn::extn_client_message::ExtnResponse,
     log::error,
     tokio::time::timeout,
+    uuid::Uuid,
 };
 
 include!(concat!(env!("OUT_DIR"), "/version.rs"));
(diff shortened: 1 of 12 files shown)


core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -115,68 +119,115 @@ pub struct AppManagerState {
     // This is a map <app_id, app_title>
     app_title: Arc<RwLock<HashMap<String, String>>>,
     app_title_persist_path: String,
+    // This is a map <app_id, app_migrated_state>
+    migrated_apps: Arc<RwLock<HashMap<String, Vec<String>>>>,
```
Vinodsathyaseelan, line 123: Why do we need to store this list in memory ?

core/sdk/src/api/storage_property.rs
```diff
@@ -547,6 +548,11 @@ impl StorageProperty {
             StorageProperty::AudioDescriptionEnabled => PROPERTY_AUDIO_DESCRIPTION_ENABLED,
             StorageProperty::PreferredAudioLanguages => PROPERTY_PREFERRED_AUDIO_LANGUAGES,
             StorageProperty::CCPreferredLanguages => PROPERTY_CC_PREFERRED_LANGUAGES,
+            StorageProperty::CustomProperty(namespace, key) => PropertyData {
+                namespace: Box::leak(namespace.clone()),
```
Vinodsathyaseelan, line 552: Where exactly does this memory get deallocated ?

core/main/src/firebolt/handlers/device_rpc.rs
```diff
@@ -193,26 +196,58 @@ pub async fn get_device_id(state: &PlatformState) -> RpcResult<String> {
...
+    if let Ok(uid) = uid_result {
+        Ok(uid)
+    } else {
+        let app_migrated = state
+            .app_manager_state
+            .get_persisted_migrated_state_for_app_id(&app_id.clone());
+
+        if app_migrated.is_some() {
+            let uid = Uuid::new_v4().to_string();
+            StorageManager::set_string(state, property, uid.clone(), None, None).await?;
```
Vinodsathyaseelan, line 212: Is it required to pass a valid `scope` value here ?

core/main/src/service/apps/delegated_launcher_handler.rs
bsenth200: Code changes are made as per the discussion with @kpears201. Doing the same here as app_title.

core/sdk/src/api/storage_property.rs
bsenth200: @VinodsathyaseelanI used Box::leak to extend the lifetime of a String to 'static because the other storage properties are all static strings. This ensures consistency in how the properties are managed. However, I understand this approach leads to intentional memory leaks, as the memory will not be deallocated. If you have suggestions for a better approach to achieve this while avoiding memory leaks, I'd be happy to consider them.

Vinodsathyaseelan: You may need to reclaim the leaked entries from `PropertyData` back to Box right after completing the usage. 

Refer : Box::from_raw()

kpears201 (approved):

device/thunder_ripple_sdk/src/processors/thunder_persistent_store.rs
```diff
@@ -92,6 +92,7 @@ impl ThunderStorageRequestProcessor {
             json!({
                 "namespace": data.namespace,
                 "key": data.key,
+                "scope": data.scope.unwrap_or_default(),
```
kpears201, line 95: so this will do an empty string for all the None cases?
Not sure we want that, shouldn't this just leave off scope if it is None?

Vinodsathyaseelan (approved):

B-Larsen (approved):

PR #547, opened 2024-06-21
RPPL-2100: Create JQ Rules for Secure Storage API

SKumarMetro:
## What

Fix the endpoint_broker handling of error case.

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -421,12 +421,13 @@ impl BrokerOutputForwarder {
                             let request_id = rpc_request.ctx.call_id;
                             v.data.id = Some(request_id);
 
-                            let mut jv: Value = Value::Null;
-                            if v.data.result.is_some() {
-                                jv = v.data.result.clone().unwrap();
-                            } else if v.data.error.is_some() {
-                                jv = v.data.error.clone().unwrap();
-                            }
+                            let jv: Value = if let Some(result) = v.data.clone().result {
+                                result
+                            } else if let Some(error) = v.data.clone().error {
+                                error
+                            } else {
+                                Value::Null
+                            };
 
                             if is_event {
                                 if let Some(filter) = broker_request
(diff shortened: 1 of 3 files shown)


core/main/src/broker/endpoint_broker.rs
```diff
@@ -420,36 +420,44 @@ impl BrokerOutputForwarder {
...
-                                            result,
-                                            &filter,
-                                            format!("{}_event", rpc_request.ctx.method),
-                                        ) {
-                                            v.data.result = Some(r);
-                                        }
-                                    }
-                                } else if let Some(filter) = broker_request
+
+                            let jv: Value = if let Some(result) = v.data.clone().result {
```
satlead, line 424: Thunder will return the event as Params
{
method: "id.event"
params: { `payload` }
}

This will not work for all brokers

core/main/src/broker/endpoint_broker.rs
```diff
@@ -420,36 +420,44 @@ impl BrokerOutputForwarder {
...
+                                result
+                            } else if let Some(error) = v.data.clone().error {
```
satlead, line 426: Please add the error handling in ThunderBroker::update_response

fn update_response(response: &JsonRpcApiResponse) -> JsonRpcApiResponse {
        let mut new_response = response.clone();
        if response.params.is_some() {
            new_response.result = response.params.clone();
        } else if response.error.is_some() {
            new_response.error = response.error.clone();
         }
        new_response
    }

pahearn73 (approved):

B-Larsen (approved):

core/main/src/broker/endpoint_broker.rs
SKumarMetro: "new_response.error = response.error.clone();"

error is already in response

// response=JsonRpcApiResponse { jsonrpc: "2.0", id: Some(1), result: None, **error:** Some(Object {"code": Number(43)}), method: None, params: None }

Not sure you meant "new_response.**result** = response.error.clone();"
Fist I did without this, then with this.

More changes needed to return an error back to the caller in endpoint_broker.

PR #553, opened 2024-06-29
Opsmetricsv2: send counters for API calls

brendanobra:
## What

Refactoring operational metrics to be standalone + adding counters for firebolt api calls

## Why
refactoring is to simplify things a bit, and counters are to provide better (lower cardinality) ops metrics, leaning into rust' stong typing a bit more to gain more advantage over high cardinality on backend : status has been forced into a defined type (not just an aribtrary struct), and the string that was formerly holding status has been replaced with the struct and types to provide conversion to the status field.

## How
added counters
## Test

## Checklist


core/sdk/src/api/observability/operational_metrics.rs
```diff
@@ -1,9 +1,220 @@
...
 };
 
+/*
+Operational Metrics. These are metrics that are not directly related to the user's behavior. They are
+more related to the operation of the platform itself. These metrics are not sent to the BI system, and
+are only used for operational/performance measurement - timers, counters, etc -all of low cardinality.
+*/
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct Counter {
```
pahearn73, line 17: This (and Timer, etc.) seem almost the same as those defined in `core/sdk/src/api/observability/mod.rs`?

core/sdk/src/api/firebolt/fb_metrics.rs
```diff
@@ -827,6 +625,18 @@ impl MetricsContext {
             authenticated: None,
         }
     }
+
+    pub fn get_fw_version(&self) -> String {
```
satlead, line 629: Seems RDK Release specific perhaps extension might be a better fit

core/sdk/src/api/observability/operational_metrics.rs
brendanobra: thanks , will clean up

core/sdk/src/api/firebolt/fb_metrics.rs
brendanobra: do u mean the name of the function is too RDK release specific, or that this function should be in the extension?

core/sdk/src/api/observability/operational_metrics.rs
brendanobra: cleaned

core/sdk/src/api/firebolt/fb_metrics.rs
satlead: yeah terms VBN and PROD are naming conventions used by RDK Release. Feel they would change in the future impacting our Ripple core. Having them in extension before they are exported seems a better strategy

satlead (approved):
RDK release specific code needs to be cleaned up to an extension in a future change

pahearn73 (approved):

PR #554, opened 2024-07-01
Http ws nonjsonrpc support

satlead:
## What

Adds support for Rule Engine to Broker requests to Websockets and HTTP which are non json rpc.
Adds Unit test support for mocking Websockets

## Why

Each device offers so many other HTTP and WS services which are not JSON RPC based. This change opens up the possibility for Ripple to Broker requests to those services within the device.

## How

Websocket Broker and HTTP Broker is updated to handle Non Json Rpc request and responses.

## Test

How has this been tested? 
Using Unit tests and internal Websocket implementation

How can a reviewer test it?
Reviewer can run a Websocket server and add rules in mock rules json and validate

## Checklist


core/main/src/broker/websocket_broker.rs
```diff
@@ -18,79 +18,200 @@
...
+    let url = url::Url::parse(&url_path).unwrap();
+    let port = extract_tcp_port(endpoint);
+    info!("Url host str {}", url.host_str().unwrap());
+    //let tcp_url = url.host_str()
+    let tcp = loop {
+        if let Ok(v) = TcpStream::connect(&port).await {
+            break v;
+        } else {
+            error!("Broker Wait for a sec and retry {}", port);
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
Vinodsathyaseelan, line 209: What if the url is not valid or the connection request is refused by server ?

core/main/src/broker/endpoint_broker.rs
```diff
@@ -421,97 +478,107 @@ impl BrokerOutputForwarder {
...
-                                        .transform
-                                        .get_filter(super::rules_engine::RuleTransformType::Event)
-                                    {
-                                        if let Ok(r) = jq_compile(
-                                            result,
-                                            &filter,
-                                            format!("{}_event", rpc_request.ctx.method),
-                                        ) {
+
+                        // Step 1: Create the data
```
Vinodsathyaseelan, line 496: For more readable and maintainable code, it may be be helpful to move the different steps to separate functions.

core/main/src/broker/websocket_broker.rs
```diff
@@ -18,79 +18,200 @@
...
+
+impl WSNotificationBroker {
+    fn start(
+        request_c: BrokerRequest,
+        callback_c: BrokerCallback,
+        url: String,
+    ) -> mpsc::Sender<String> {
+        let (tx, mut tr) = mpsc::channel::<String>(1);
+        tokio::spawn(async move {
+            let (final_url, tcp) = connect(&url, Some(request_c.clone().rule.alias)).await;
```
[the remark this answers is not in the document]
satlead: Websocket Brokers are intended for critical internal websockets so Connect is expected to succeed we keep retrying till it does based on the current expectation. This is the current strategy for the first phase of this implementation we can make this more sophisticated only if necessary.

core/main/src/broker/websocket_broker.rs
satlead: Websocket Brokers are intended for critical internal websockets so Connect is expected to succeed we keep retrying till it does based on the current expectation. This is the current strategy for the first phase of this implementation we can make this more sophisticated only if necessary. URL is configured on the rules file like any bad configuration will have impacts. In this case that particular API call will indefinitely wait.

core/main/src/broker/websocket_broker.rs
```diff
@@ -18,79 +18,200 @@
...
+                let (mut ws_tx, mut ws_rx) = stream.split();
+
+                tokio::pin! {
+                    let read = ws_rx.next();
+                }
+                loop {
+                    tokio::select! {
+                        Some(value) = &mut read => {
+                            match value {
+                                Ok(v) => {
```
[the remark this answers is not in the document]
satlead: Fragmentation is probably on the lower TCP level , given Websockets deal with messages entire fragments get recreated before getting passed to the sink. Also given we are async smaller messages might make it before bigger messages which should be fine as well in this new architecture

core/main/src/broker/websocket_broker.rs
```diff
@@ -18,79 +18,200 @@
...
 }
 
+async fn connect(endpoint: &str, alias: Option<String>) -> (url::Url, TcpStream) {
+    info!("Broker Endpoint url {}", endpoint);
+    let url_path = if let Some(a) = alias {
+        format!("{}{}", endpoint, a)
+    } else {
+        endpoint.to_owned()
+    };
+    let url = url::Url::parse(&url_path).unwrap();
```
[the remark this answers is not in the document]
satlead: There is no fallback strategy for badly configured Websocket brokers this includes Thunder Broker too. Unsafe unwrap might not be cleaner but for a first phase feel this is adequate as the errors will be very fatal during startup.

We can plan for some graceful shutdown on a future ticket.

core/main/src/broker/endpoint_broker.rs
satlead: This method was already refactored challenge was sharing the variables across these refactored methods which made it very. Did little bit more refactoring anything more would increase the parameter complexity of these methods

Vinodsathyaseelan (approved):

Nv-Delle1416 (approved):

PR #561, opened 2024-07-10
RPPL-2360:Update LOG level for ripple.service

pkumbh631:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/systemd/ripple.service
+++ b/systemd/ripple.service
@@ -4,8 +4,8 @@ After=wpeframework.service
 
 [Service]
 Type=notify
-Environment="RUST_LOG=debug"
 EnvironmentFile=-/opt/ripple.conf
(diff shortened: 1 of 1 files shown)


B-Larsen (approved):

systemd/ripple.service
```diff
@@ -4,8 +4,8 @@ After=wpeframework.service
 
 [Service]
 Type=notify
-Environment="RUST_LOG=debug"
```
brendanobra, line 7: IMHO, this should still be here as a fallback of last resort, as `Environment="RUST_LOG=info"`

satlead: Would System Services set the Environment values before ExecStartPre

brendanobra: Seems like it , I was wondering if we needed to keep this as a fallback in case something happens in `ExecStartPre`.

brendanobra (approved):

PR #567, opened 2024-07-15
Support for x-provided-by Pull pattern

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/provider_registrar.rs
+++ b/core/main/src/firebolt/handlers/provider_registrar.rs
@@ -15,13 +15,18 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use std::time::Duration;
+
 use crate::{
     firebolt::rpc::register_aliases,
-    service::apps::provider_broker::ProviderBroker,
-    state::{cap, platform_state::PlatformState},
+    service::apps::{
+        app_events::AppEvents,
+        provider_broker::{ProviderBroker, ProviderBrokerRequest},
+    },
+    state::{cap, openrpc_state::ProviderSet, platform_state::PlatformState},
 };
 use jsonrpsee::{
-    core::{server::rpc_module::Methods, RpcResult},
+    core::{server::rpc_module::Methods, Error, RpcResult},
     types::{error::CallError, ParamsSequence},
     RpcModule,
 };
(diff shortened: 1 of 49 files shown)


core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -103,138 +132,610 @@ impl ProviderRegistrar {
...
-
-                                    Ok(ListenerResponse {
-                                        listening,
-                                        event: attributes.event.into(),
-                                    })
-                                },
-                            )
-                            .unwrap();
-                    }
+    fn register_method(
```
pahearn73, line 135: Registers firebolt method with jsonrpsee.

core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -103,138 +132,610 @@ impl ProviderRegistrar {
...
+
+        match result {
+            Ok(_) => true,
+            Err(e) => {
+                error!("register_method: Error registering method: {:?}", e);
+                false
+            }
+        }
+    }
+
```
pahearn73, line 177: Various callback types follow (`callback_*()`), these are effectively executed when the associated firebolt API call is made.

core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -103,138 +132,610 @@ impl ProviderRegistrar {
...
+            return Err(Error::Custom(String::from("Missing provider attributes")));
+        }
+
+        Ok(None) as RpcResult<Option<()>>
+    }
+
+    async fn callback_provider_invoker(
+        params: Params<'static>,
+        context: Arc<RpcModuleContext>,
+    ) -> Result<Value, Error> {
```
pahearn73, line 276: I'm not too sure about this. If we want to avoid mapping response data to specific types this needs to return a  `Value` so the receiver of the response can deserialize it as it knows the type. But does this clash with the way the firebolt schema is currently defined? Not sure how we want to handle this.

core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -103,138 +132,610 @@ impl ProviderRegistrar {
...
-                                    }
-
-                                    Ok(None) as RpcResult<Option<()>>
-                                },
-                            )
-                            .unwrap();
+        Ok(None)
+    }
+
+    pub fn register_methods(platform_state: &PlatformState, methods: &mut Methods) -> u32 {
```
pahearn73, line 391: Iterates though the provider map and registers firebolt methods by type.

core/main/src/state/cap/generic_cap_state.rs
```diff
@@ -49,6 +49,7 @@ impl GenericCapState {
             FireboltCap::Short("token:platform".to_owned()),
             FireboltCap::Short("usergrant:acknowledgechallenge".to_owned()),
             FireboltCap::Short("usergrant:pinchallenge".to_owned()),
+            // <pca> Do we need to add "discovery:interest" etc. here or just in the manifests? </pca>
```
pahearn73, line 52: I'm not sure why these are hardcoded here, I think when `GenericCapState::new()` is called the manifest has already been ingested so what's the point, why not just include these in the manifest?

satlead: When Ripple starts up these capabilities are set to unavailable. So any calls for these providers would fail with unavailable error. 

Once the providers are registered the broker would then set these capabilities as available. From that point the consecutive calls would pass

core/main/src/state/firebolt-open-rpc.json
satlead, line 1: @rdkcentral/ripple-qa please note

core/main/src/firebolt/handlers/provider_registrar.rs
satlead: Feel this could be a Value as long as we evaluate using the OpenRpc Validator for Request/Response and Event. Making Ripple an effective Pass through with Schema validation support,

satlead (approved):
LGTM

core/main/src/state/cap/generic_cap_state.rs
pahearn73: @satlead: Right but these get set as a result of `BootstrapState::build()` being called, which is before the providers are registered. Because of that they might as well just be defined in the manifest right?

core/main/src/firebolt/handlers/provider_registrar.rs
pahearn73: I think calls like `Content.requestUserInterest` expect the actual struct to be returned, e.g. in this case `EntityDetails` and not `Value::Object(EntityDetails)`, is that right @kpears201 ?

pahearn73: ...and if that's the case how will we ever get away from having to map response types like we still do in `ProviderRegistrar::get_provider_reponse()` ([here](https://github.com/rdkcentral/Ripple/blob/main/core/main/src/firebolt/handlers/provider_registrar.rs#L57))?

pahearn73: Another point, for responses from APIs that have `x-provided-by`, ripple needs to be able to set the app ID parameter to the value of the app ID parameter (if it exists) in the provider response ([RPPL-2310](https://ccp.sys.comcast.net/browse/RPPL-2310)), meaning ripple needs to be able to process the response and not just pass-through.

Nv-Delle1416 (approved):

PR #568, opened 2024-07-15
RPPL-2148 [Storage Broker] Cleanup Localization

sakshihcst:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/rules_engine.rs
+++ b/core/main/src/broker/rules_engine.rs
@@ -37,7 +37,15 @@ pub struct RuleSet {
 impl RuleSet {
     pub fn append(&mut self, rule_set: RuleSet) {
         self.endpoints.extend(rule_set.endpoints);
-        self.rules.extend(rule_set.rules);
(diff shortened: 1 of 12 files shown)


core/main/src/firebolt/handlers/localization_rpc.rs
```diff
@@ -183,6 +176,35 @@ pub trait Localization {
     ) -> RpcResult<ListenerResponse>;
 }
 
+pub async fn get_language(ctx: &CallContext, state: &PlatformState) -> RpcResult<String> {
```
satlead, line 179: Why do we need this if there is a rule?

Vinodsathyaseelan (approved):

Vinodsathyaseelan (changes requested):
The usage of `main_internal_request()` call  for `localization.language` from `LoadDistributorValuesStep::setup() -> MetricsState::initialize()`  will cause bootstrap lock up (indefinite wait).  Covert this PR to draft for re-work.

PR #575, opened 2024-07-19
fix ontology device_type is not reported correctly

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/state/metrics_state.rs
+++ b/core/main/src/state/metrics_state.rs
@@ -61,7 +61,7 @@ const PERSISTENT_STORAGE_ACCOUNT_DEVICE_MANUFACTURER: &str = "deviceManufacturer
 pub enum SiftDeviceType {
     IPTV,
     IPSTB,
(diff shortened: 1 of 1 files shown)


core/main/src/state/metrics_state.rs
```diff
@@ -342,7 +357,14 @@ impl MetricsState {
                 .await
             {
                 Some(s) => s,
-                None => state.get_device_manifest().get_form_factor(),
+                None => {
+                    let form_factor = state.get_device_manifest().get_form_factor();
```
pahearn73, line 361: Oops, I think I gave you bad advice on doing the mapping here, because we can't have any SIFT-related stuff in the `Ripple` repo... I think those references need to go back in `ripple-eos-distributor-extn` where you had them, maybe do the mapping when the `OntologyPayload` is assembled.

tfarz123 (approved):

PR #577, opened 2024-07-23
fix: Bug fix in on_status_response() function

Vinodsathyaseelan:
## What

Bug fix 

## Why

This fix is required to process all pending broker requests upon receiving response of plugin status command, irrespective of the activation status

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/thunder/thunder_plugins_status_mgr.rs
+++ b/core/main/src/broker/thunder/thunder_plugins_status_mgr.rs
@@ -22,7 +22,7 @@ use std::{
 use ripple_sdk::{
     api::gateway::rpc_gateway_api::JsonRpcApiResponse,
     chrono::{DateTime, Duration, Utc},
-    log::{info, warn, error},
+    log::{error, info, warn},
     utils::error::RippleError,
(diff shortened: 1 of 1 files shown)


core/main/src/broker/thunder/thunder_plugins_status_mgr.rs
```diff
@@ -416,19 +416,17 @@ impl StatusManager {
...
-                            .await;
-                    } else {
-                        let _ = sender.send(pending_request).await;
-                    }
+            let (pending_requests, expired) =
+                self.retrive_pending_broker_requests(callsign.to_string());
+
+            for pending_request in pending_requests {
+                if expired {
+                    info!("Expired request: {:?}", pending_request);
```
brendanobra, line 424: I realize that this is from the original code but: does this need to be info!? maybe debug is better?

satlead: +1

satlead (approved):

Nv-Delle1416 (approved):

PR #582, opened 2024-07-27
fix JQ parser on filter that extracting json object

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/rules_engine.rs
+++ b/core/main/src/broker/rules_engine.rs
@@ -176,15 +176,16 @@ pub fn jq_compile(input: Value, filter: &str, reference: String) -> Result<Value
     // start out only from core filters,
     // which do not include filters in the standard library
     // such as `map`, `select` etc.
-    let mut defs = ParseCtx::new(Vec::new());
 
+    let mut defs = ParseCtx::new(vec!["fromjson".into()]);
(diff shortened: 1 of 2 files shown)


core/main/src/broker/rules_engine.rs
```diff
@@ -176,15 +176,16 @@ pub fn jq_compile(input: Value, filter: &str, reference: String) -> Result<Value
     // start out only from core filters,
     // which do not include filters in the standard library
     // such as `map`, `select` etc.
-    let mut defs = ParseCtx::new(Vec::new());
 
+    let mut defs = ParseCtx::new(vec!["fromjson".into()]);
```
Vinodsathyaseelan, line 180: Do we need to keep adding other commands in this list ?  If yes, it would be better to move this on a configuration file.

Vinodsathyaseelan (approved):

core/main/src/broker/rules_engine.rs
satlead: +1 for configuration we need lots of other definitions to be supported as well

satlead (approved):

B-Larsen (approved):

PR #589, opened 2024-08-05
WIP: jq rules engine refactor/tests

brendanobra:
## What

1) ripple core/main as a library (ostensibly to enabled automated testing of jq rules)
2) Rules engine refactor to a) decouple rule processing workflow from tokio b) make smaller things that are more easily reasoned about, are easier to unit test
3) Explicit/deterministic error reporting/handling for jq rules
4) Unit test related changes a) defaults for many oft used classes b) unit tests for broker + rules
5) Added `Default` implementation (both explicit and derived , where possible)
## Why

Automated testing of jq rules is currently not possible
Errors are obfuscated.

## How

Refactor to expose appropriate components, more idiomatic use of rust' type system

## Test

`cargo test`

## Checklist

- [ X] I have self-reviewed this PR
- [ X]   I have added tests that prove the feature works or the fix is effective

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -26,7 +26,7 @@ use ripple_sdk::{
     },
     extn::extn_client_message::{ExtnEvent, ExtnMessage},
     framework::RippleResponse,
-    log::error,
+    log::{debug, error, info},
     semver::Op,
     tokio::{
         self,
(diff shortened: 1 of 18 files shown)


core/main/src/broker/endpoint_broker.rs
```diff
@@ -187,22 +194,40 @@ pub struct BrokerContext {
...
+                params: None,
+            },
+        }
+    }
+}
 
+pub fn get_event_id_from_method(method: Option<String>) -> Option<u64> {
+    method.and_then(|m| {
+        let event: Vec<&str> = m.split('.').collect();
+        event.first().and_then(|v| v.parse::<u64>().ok())
```
satlead, line 215: Add some safety for bad methods

core/main/src/broker/endpoint_broker.rs
```diff
@@ -502,7 +540,7 @@ pub trait EndpointBroker {
     }
 
     /// Generic method which takes the given parameters from RPC request and adds rules using rule engine
-    fn apply_request_rule(rpc_request: &BrokerRequest) -> Result<Value, RippleError> {
+    fn apply_request_rule(rpc_request: &BrokerRequest) -> Result<Value, JqError> {
```
satlead, line 543: Can we keep them as Ripple Error instead of JqError

Ripple error is already an enum which does the same for Deny Reason

core/main/src/broker/endpoint_broker.rs
```diff
@@ -545,90 +583,206 @@ pub trait EndpointBroker {
...
+}
+
+/*
+
+Factor out broker workflow from tokio loop
+*/
+pub fn run_broker_workflow(
+    broker_output: &BrokerOutput,
+    broker_request: &BrokerRequest,
+) -> Result<BrokerWorkflowSuccess, BrokerWorkFlowError> {
```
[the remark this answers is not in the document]
brendanobra: Same as above, the goal of this approach is to provide more information, with a better chance of figuring out where the error was generated.  What

core/main/src/broker/endpoint_broker.rs
brendanobra: RippleError is overloaded imho, which is why i leaned into this more rust idiomatic apporach.

PR #590, opened 2024-08-06
Audio  descriptions + closed captions -> jq rules

brendanobra:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -25,7 +25,7 @@ use ripple_sdk::{
     },
     extn::extn_client_message::{ExtnEvent, ExtnMessage},
     framework::RippleResponse,
-    log::error,
+    log::{debug, error},
     tokio::{
         self,
         sync::mpsc::{self, Receiver, Sender},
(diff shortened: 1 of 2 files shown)


core/main/src/broker/endpoint_broker.rs
```diff
@@ -204,6 +209,12 @@ impl BrokerOutput {
         }
         None
     }
+    pub fn is_success(&self) -> bool {
+        self.data.error.is_none()
```
Vinodsathyaseelan, line 213: Is it better to validate the `result` field here ?

core/main/src/broker/endpoint_broker.rs
```diff
@@ -204,6 +209,12 @@ impl BrokerOutput {
...
+    }
+    pub fn is_failure(&self) -> bool {
+        !self.is_success()
```
Vinodsathyaseelan, line 216: Similar comment, if `error` field has some value, then data contains error response.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -685,24 +692,69 @@ async fn forward_extn_event(
...
+                            } else {
+                                None
+                            }
+                        } else {
+                            None
+                        }
+                    }
+                    None => None,
+                };
+            } else if raw_value.get("success").is_some() {
```
Vinodsathyaseelan, line 734: This needs to be re-checked for `result` and `error` use cases.    
`result` response struct is implementation defined and `success` field is not part of standard JSON-RPC 2.0 response format.

PR #591, opened 2024-08-07
fix endpoint broker apply response parse json rpc response

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -651,17 +651,7 @@ impl BrokerOutputForwarder {
                                     super::rules_engine::RuleTransformType::Response,
                                 )
                             {
-                                let error_response_filter =
-                                    broker_request.rule.transform.get_filter(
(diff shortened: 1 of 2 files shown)


satlead (approved):

core/main/src/broker/endpoint_broker.rs
```diff
@@ -686,6 +686,10 @@ async fn forward_extn_event(
 }
 
 fn apply_response(result: Value, filter: String, rpc_request: &RpcRequest, v: &mut BrokerOutput) {
+    if result.is_null() {
```
Vinodsathyaseelan, line 689: apply_response is called by passing only the `result` field of `JsonRpcApiResponse`.   A `None`  value for this field indicates an `error` response from the RPC call.  

Where are we processing the error response?   Looks like this change will mask `v.data.error = Some(r);`

brendanobra (approved):

Vinodsathyaseelan (approved):

Nv-Delle1416 (approved):

PR #599, opened 2024-08-09
feat: Configurable Provider Registrations

satlead:
## What

Provider registrations are hard coded in ripple, making it configurable.

## Why

Hard coded configurations are not scalable for extension sdks.

## How

Having a default set of configurations and allow it to be overridden by manifests

## Test

Test without any manifest changes and make sure all the provider pattern works
Add an entry in provider_registrations for extension manifest and check if other providers fail with method not found.

## Checklist

### DIFF
--- a/core/main/src/state/openrpc_state.rs
+++ b/core/main/src/state/openrpc_state.rs
@@ -67,7 +67,7 @@ impl ProviderRelationSet {
 fn is_provider_enabled(method: &str, checksets: &Vec<String>) -> bool {
     let method_lower_case = method.to_lowercase();
     for checkset in checksets {
-        if checkset.to_lowercase().eq(&method_lower_case) {
+        if method_lower_case.starts_with(&checkset.to_lowercase()) {
             return true;
         }
     }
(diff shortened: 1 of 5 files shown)


core/main/src/state/openrpc_state.rs
```diff
@@ -64,27 +64,29 @@ impl ProviderRelationSet {
...
+        if method_lower_case.starts_with(&checkset.to_lowercase()) {
+            return true;
+        }
+    }
+    false
+}
+
 pub fn build_provider_relation_sets(
     openrpc_methods: &Vec<FireboltOpenRpcMethod>,
+    checksets: &Vec<String>,
```
pahearn73, line 79: Since you've added the new field `provider_registrations: Vec<String>` to `OpenRpcState`, can we please make both `build_provider_relation_sets` and `is_provider_enabled` `OpenRpcState` methods instead of functions (e.g. `build_provider_relation_sets(&self...`), `is_provider_enabled(&self...`) so that we don't have to pass around the provider list?

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #606, opened 2024-08-20
fixes: http_broker bugfixes

brendanobra:
## What

for http_broker:
1) fixed bugs that were 1) not using uri in config 2) only support POST (it now only supports GET, which is all that is needed presently) 3) not using the rules engine 4) general refactor to be a bit more idiomatic/readable

## Why
http_broker was not working

## How

fixes/improvements implemented
## Test
Tested against hw
## Checklist

- [ x] I have self-reviewed this PR
(unit tests will follow time permitting)


core/main/src/broker/http_broker.rs
```diff
@@ -15,78 +15,143 @@
...
+                match send_http_request(&client, Method::GET, &uri, &request.clone().rule.alias)
+                    .await
+                {
+                    Ok(response) => {
+                        let (parts, body) = response.into_parts();
+
+                        let body = body_to_bytes(body).await;
+                        let mut request = request;
+
+                        let v = vec![serde_json::from_slice::<serde_json::Value>(&body).unwrap()];
```
pahearn73, line 130: Should we be concerned about unwrapping response data from external sources?

brendanobra: We definitely should, and now are :) - Thanks!

pahearn73 (approved):

maggie98choy (approved):

Nv-Delle1416 (approved):

PR #607, opened 2024-08-21
RPPL-2452: Firebolt Ripple JQ Rules: Support for static response data

SKumarMetro:
## What

Firebolt Ripple JQ Rules: Support for static response data

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -466,7 +466,6 @@ impl EndpointBrokerState {
         let mut broker_sender = None;
         let mut found_rule = None;
         if let Some(rule) = self.rule_engine.get_rule(&rpc_request) {
-            // ==> handle_brokerage: Rule { alias: "static", transform: RuleTransform { request: None, response: Some("WPE"), event: None, event_decorator_method: None }, filter: None, endpoint: None }
             let _ = found_rule.insert(rule.clone());
             if let Some(endpoint) = rule.endpoint {
                 if let Some(endpoint) = self.get_sender(&endpoint) {
@@ -489,7 +488,6 @@ impl EndpointBrokerState {
             let updated_request = self.update_request(&rpc_request, rule, extn_message);
             tokio::spawn(async move {
                 if let Err(e) = broker.send(updated_request.clone()).await {
-                    // send some rpc error
                     callback.send_error(updated_request, e).await
                 }
             });
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -442,8 +442,10 @@ impl EndpointBrokerState {
                 if let Some(endpoint) = self.get_sender(&endpoint) {
                     let _ = broker_sender.insert(endpoint);
                 }
-            } else if let Some(endpoint) = self.get_sender("thunder") {
-                let _ = broker_sender.insert(endpoint);
+            } else if rule.alias != "static"  {
+                if  let Some(endpoint) = self.get_sender("thunder") {
+                    let _ = broker_sender.insert(endpoint);
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


core/main/src/broker/endpoint_broker.rs
```diff
@@ -404,6 +414,43 @@ impl EndpointBrokerState {
         }
     }
 
+    fn handle_static_ruquest(
```
brendanobra, line 417: nitpick: `request`

core/main/src/broker/endpoint_broker.rs
```diff
@@ -404,6 +414,43 @@ impl EndpointBrokerState {
...
+        &self,
+        rpc_request: RpcRequest,
+        extn_message: Option<ExtnMessage>,
+        rule: Rule,
+        callback: BrokerCallback,
+    ) {
+        let mut data = JsonRpcApiResponse::default();
```
brendanobra, line 424: can instansiation of `data` be moved inside of `let` to avoid creating when the let does not match?

core/main/src/broker/endpoint_broker.rs
SKumarMetro: done

core/main/src/broker/endpoint_broker.rs
SKumarMetro: done

core/main/src/broker/endpoint_broker.rs
```diff
@@ -326,6 +326,16 @@ impl EndpointBrokerState {
         rule: Rule,
         extn_message: Option<ExtnMessage>,
     ) -> BrokerRequest {
+        let (_id, req) = self.update_request2(rpc_request, rule, extn_message);
+        req
+    }
+
+    fn update_request2(
```
pahearn73, line 333: This is virtually identical to `update_request()` [here](https://github.com/rdkcentral/Ripple/blob/7f3ebb7d4bd2708499f8b93da7bddf99a3518861/core/main/src/broker/endpoint_broker.rs#L323), can we just modify the existing method as necessary?

core/main/src/broker/endpoint_broker.rs
```diff
@@ -433,14 +462,18 @@ impl EndpointBrokerState {
...
-        let updated_request = self.update_request(&rpc_request, rule, extn_message);
-        tokio::spawn(async move {
-            if let Err(e) = broker.send(updated_request.clone()).await {
-                // send some rpc error
-                callback.send_error(updated_request, e).await
-            }
-        });
+
+        if rule.alias == "static" {
+            self.handle_static_request(rpc_request, extn_message, rule, callback);
```
pahearn73, line 467: I think we want to handle static rules before we check if `broker_sender.is_none()`, because static rules will not have a broker right?

core/main/src/broker/endpoint_broker.rs
```diff
@@ -404,6 +414,25 @@ impl EndpointBrokerState {
...
     }
 
+    fn handle_static_request(
+        &self,
+        rpc_request: RpcRequest,
+        extn_message: Option<ExtnMessage>,
+        rule: Rule,
+        callback: BrokerCallback,
+    ) {
+        if rule.transform.response.is_some() {
```
pahearn73, line 424: `transform.response` is required for any rule that has a static alias, so I'm not really sure we need this check.

core/main/src/broker/endpoint_broker.rs
SKumarMetro: Fixed

core/main/src/broker/endpoint_broker.rs
SKumarMetro: No point in having static rule with response, that's why there was a check.  
Removed the check, static rule without response will return 'result: ""'

core/main/src/broker/endpoint_broker.rs
SKumarMetro: Rearranged the conditions.

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #610, opened 2024-08-22
eos-ripple: run.sh no longer updates firebolt-app-library.json location

pkumbh631:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/bootstrap/manifest/apps.rs
```diff
@@ -27,6 +27,36 @@ pub struct LoadAppLibraryStep;
 
 impl LoadAppLibraryStep {
     pub fn load_app_library(path: String) -> Result<Vec<AppLibraryEntry>, RippleError> {
+        // Check if the "local_dev" feature is enabled
```
pahearn73, line 30: For both the device manifest and extension manifest we try and load the files from various sources depending on whether or not the previous source failed, environment variables set, etc. For example, if `local_dev` is enabled we first try to load based on the environment variable and if that fails we fall back to the home directory, [here](https://github.com/rdkcentral/Ripple/blob/main/core/main/src/bootstrap/manifest/device.rs#L37). For consistency we should probably follow the same exact pattern (load_from_home/load_from_env/load_from_etc) by copying `try_manifest_files()` and related functions. Just duplicate what we do for device and extension manifests.

pahearn73 (approved):

dhillomk (approved):

PR #612, opened 2024-08-23
Sync'd 1.11.rc --> 1.12.rc

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -787,6 +787,7 @@ fn apply_response(
                     error!("jq_compile error {:?}", e);
                 }
             }
+            trace!("mutated output {:?}", v);
         }
         Err(e) => {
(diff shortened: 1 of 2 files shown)


core/main/src/broker/endpoint_broker.rs
```diff
@@ -787,6 +787,7 @@ fn apply_response(
                     error!("jq_compile error {:?}", e);
                 }
             }
+            trace!("mutated output {:?}", v);
```
B-Larsen, line 790: @pahearn73 It looks like this was removed under https://github.com/rdkcentral/Ripple/commit/5e523501d91d9890a9a57767ecd53f248c0a1b5f
and added in line 783.  Is it intentional to add this back in here as well?

pahearn73:
Replacing with pahearn_1.11.rc-1.12.rc_b

PR #615, opened 2024-08-28
fix: Handling for Provider Error for generic

satlead:
## What

Currently Provider errors are not handled for Providers outside Pin Challenge and Acknowledge Challenge

## Why

Original changes were focused on the Response payloads.

## How

By providing a Generic Error Payload as a default parser for any Provider Responses before throwing errors.

## Test

Using Error Type object for Any API call outside Acknowledge Challenge and Pin Challenge

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/provider_registrar.rs
+++ b/core/main/src/firebolt/handlers/provider_registrar.rs
@@ -133,13 +133,17 @@ impl ProviderRegistrar {
                 }
             }
             ProviderResponsePayloadType::GenericResponse => {
-                let external_provider_response: Result<ExternalProviderResponse<Value>, CallError> =
-                    params_sequence.next();
+                let external_provider_response: Result<
+                    ExternalProviderResponse<Option<Value>>,
+                    CallError,
+                > = params_sequence.next();
 
                 if let Ok(r) = external_provider_response {
                     return Some(ProviderResponse {
                         correlation_id: r.correlation_id,
-                        result: ProviderResponsePayload::GenericResponse(r.result),
+                        result: ProviderResponsePayload::GenericResponse(
+                            r.result.unwrap_or(Value::Null),
+                        ),
                     });
                 }
             }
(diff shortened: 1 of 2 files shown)


core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -119,15 +119,18 @@ impl ProviderRegistrar {
...
-                if let Ok(r) = external_provider_error {
-                    return Some(ProviderResponse {
-                        correlation_id: r.correlation_id,
-                        result: ProviderResponsePayload::ChallengeError(r.error),
-                    });
+                match external_provider_error {
+                    Ok(r) => {
+                        return Some(ProviderResponse {
+                            correlation_id: r.correlation_id,
+                            result: ProviderResponsePayload::ChallengeError(r.error),
```
pahearn73, line 130: Are you sure this is always going to be a `ChallengeError` when the payload type is `ProviderResponsePayloadType::GenericError`?

core/sdk/src/api/firebolt/provider.rs
```diff
@@ -49,6 +49,7 @@ pub enum ProviderResponsePayloadType {
     EntityInfoResponse,
     PurchasedContentResponse,
     Generic,
```
pahearn73, line 51: Since we've added `GenericError` let's rename `Generic` --> `GenericResponse`.

core/main/src/firebolt/handlers/provider_registrar.rs
satlead: So firebolt open rpc has a generic provider for all error objects which match the ChallengeError 
https://github.com/rdkcentral/firebolt-openrpc/blob/665aad3ece7e8b3e196c1a1bafa34181c646c26e/src/shared/modules.mjs#L878

So i just renamed the challenge Error to generic error now its very scalable

pahearn73: Nice, thanks.

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #622, opened 2024-09-04
RPPL-2419: deleted the firebolt-open-rpc.json from Ripple, necessary erro…

nnaveen979:
…r handling updated.

## What

Remove: firebolt-open-rpc.json file from Ripple.

## Why

This will reduce the ripple filesystem footprint by approx 257kb-501kb

## How

deleted the firebolt-open-rpc.json from Ripple, by environment variables , build system still access this file. full details updated in Ripple README file.

## Test

firebolt-open-rpc.json file deleted and run Ripple, it's referring the expected path.

## Checklist

- [+] I have self-reviewed this PR


pahearn73 (commented):
We should be attempting to load firebolt-open-rpc.json exactly the same way we do for manifests, e.g. for device manifest we have try_manifest_files() which tries to load from various sources e.g. load_from_env(), load_from_home(), load_from_etc(). Create these same functions, do it the same way. See [here](https://github.com/rdkcentral/Ripple/blob/main/core/main/src/bootstrap/manifest/device.rs#L25).

Your unit tests are failing because the path to firebolt-open-rpc.json is empty. That's probably because you haven't added:

export FIREBOLT_OPEN_RPC=${workspace_dir}/target/openrpc/firebolt-open-rpc.json

...to run.sh and runCI.sh.

core/main/src/state/openrpc_state.rs
```diff
@@ -125,55 +125,20 @@ impl OpenRpcState {
         }
     }
 
-    fn load_firebolt_open_rpc() -> FireboltVersionManifest {
```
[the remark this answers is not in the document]
brendanobra: run.ch and runCI.sh are not used in the unit test step in the github action that runs the ci for PRs (https://github.com/rdkcentral/Ripple/blob/main/.github/workflows/Build.yml) so this ^^^ will not work. The issue appears to be that  that there is a coupling between the unit tests and `firebolt-open-rpc.json`, which does not exist in Ripple (and only seems to exist in a closed source repo). I think the best way to proceed is to either 1) add a `firebolt-open-rpc.json` into Ripple (which may have to be maintained, not sure of the details) or 2) add mock/test gated compilation for different behavior during test in openrpc_state.rs (this is probably better, but I have not looked too deeply into it yet)

.github/workflows/Build.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions-rust-lang/setup-rust-toolchain@v1
-      - run: cargo test
+      - run: ./ripple test
```
brendanobra, line 44: As we discussed, I don't think this is best practice  - the unit tests should run with only `cargo test`, and not require a wrapper to setup the environment

nnaveen979: I agree, will remove these changes.

pahearn73 (approved):

brendanobra (approved):

tfarz123 (approved):

PR #631, opened 2024-09-19
feat: HDCP/HDR settings cache updated

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/client/thunder_client.rs
+++ b/device/thunder_ripple_sdk/src/client/thunder_client.rs
@@ -394,11 +394,11 @@ impl ThunderClient {
                 if let Some(sub) = subscriptions.get_mut(&subscribe_method) {
                     // Remove the listener for the given sub_id
                     sub.listeners.remove(&sub_id);
-                    
-                    sub.sub_ids_list.get_mut(&subscribe_method).map(|sub_ids| {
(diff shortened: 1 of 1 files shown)


core/main/src/service/context_manager.rs
```diff
@@ -75,6 +75,34 @@ impl ContextManager {
             warn!("No processor to set TimeZoneChanged status listener")
         }
 
+        // Setup the DisplayConnectionChanged status listener
+        debug!("Subscribing for change in Display Connection");
+        if ps
+            .get_client()
+            .send_extn_request(DeviceEventRequest {
+                event: DeviceEvent::InputChanged,
```
pahearn73, line 83: I'm not sure we should be using `DeviceEvent::InputChanged` at all with any of this, I think this refers to HDMI inputs whereas HDCP relates to when the OUTPUT device (TV) is changed (e.g. device connection has changed).

pahearn73: ...in other words it seems like we need to rename `DeviceEvent::InputChanged` to `DeviceEvent::HdcpChanged`. But note the contract name should also be changed (`EventAdjective::Input` --> `EventAdjective::Hdcp`) in code and manifests.

pahearn73 (dismissed):

PR #633, opened 2024-09-20
RPPL-2593 Device.uid is returning a value that contains dash

SKumarMetro:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/rpc_router.rs
+++ b/core/main/src/firebolt/rpc_router.rs
@@ -134,15 +134,23 @@ async fn resolve_route(
 impl RpcRouter {
     pub async fn route(
         state: PlatformState,
-        req: RpcRequest,
+        mut req: RpcRequest,
         session: Session,
         timer: Option<Timer>,
     ) {
         let methods = state.router_state.get_methods();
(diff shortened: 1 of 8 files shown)


core/main/src/firebolt/rpc_router.rs
```diff
@@ -134,15 +134,23 @@ async fn resolve_route(
...
         let methods = state.router_state.get_methods();
         let resources = state.router_state.resources.clone();
 
+        let method = req.method.clone();
+        if let Some(overriden_method) = state
+            .get_manifest()
+            .has_rpc_override_method(&req.method)
+        {
+            req.method = overriden_method;
+            println!("route: overridding ")
```
satlead, line 150: Remove the println we can probably add a proper debug here

satlead (approved):

core/main/src/firebolt/rpc_router.rs
SKumarMetro: Removed

satlead (approved):

B-Larsen (approved):

PR #642, opened 2024-10-01
Jq rule composition - workflow for async getters

brendanobra:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/endpoint_broker.rs
```diff
@@ -15,11 +15,13 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use jsonrpsee::tracing::debug;
```
Vinodsathyaseelan, line 18: `ripple_sdk::log::debug` here ?

pahearn73 (approved):

core/main/src/broker/workflow_broker.rs
```diff
@@ -0,0 +1,316 @@
...
+                            "alias": "static",
+                            "transform": {
+                              "response": "\"Sky\""
+                            }
+                        },
+                        "module.method": {
+                            "alias": "workflow",
+                            "endpoint": "workflow",
+                            "sources": [{
+                                "method": "static.rle"
```
satlead, line 286: there is a typo here for static.rle does it still work is the test for invalid rule?

core/sdk/src/api/gateway/rpc_gateway_api.rs
```diff
@@ -197,6 +198,60 @@ impl JsonRpcApiRequest {
     }
 }
 
+#[derive(Clone, Default, Debug)]
+pub struct JsonRpcApiError {
```
satlead, line 202: Was this added because JsonRPCAPIResponse doesnt have `message:String` Just curious why we need this new entry?

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct JsonRpcApiResponse {
    pub jsonrpc: String,
    pub id: Option<u64>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub result: Option<Value>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub error: Option<Value>,
    #[serde(skip_serializing)]
    pub method: Option<String>,
    #[serde(skip_serializing)]
    pub params: Option<Value>,
}

brendanobra: This (JsonApiError)was added to add stronger typing (as part of my original POC). This will (does) ultimately make code more understandable and enable proper encapsulation, higher type safety and ultimately better maintenance and unit testing in the future.

core/main/src/broker/workflow_broker.rs
brendanobra: In this case (the unit test test RuleEngine::load_from_string_literal, which makinly tests that it can take a schema compliant string as input, and produce a rule engine from them), so it does not make any difference (but does look a little confusing now that i'm looking at it)

core/main/src/bootstrap/start_communication_broker.rs
```diff
@@ -68,7 +68,7 @@ impl Bootstep<BootstrapState> for StartOtherBrokers {
         let session = ps.session_state.get_account_session();
         // Setup the endpoints from the manifests
         let mut endpoint_state = ps.endpoint_state;
-        endpoint_state.build_other_endpoints(session);
+        endpoint_state.build_other_endpoints(session.clone());
```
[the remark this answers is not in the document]
brendanobra: could probably even do this:
  endpoint_state.build_other_endpoints( ps.session_state.get_account_session());

satlead (approved):

Nv-Delle1416 (approved):

PR #647, opened 2024-10-08
RPPL-2651: Ripple Event warning should not print for context events

nnaveen979:
## What

add check for event listener's for context events.

## Why

if there are no event listeners available then no need to warn

## How

added check for event listeners availability based on that print the warn!

## Test

if context event listeners available then warn! msg prints if not then warn! msg wont print.

## Checklist

- [+] I have self-reviewed this PR

### DIFF
--- a/core/sdk/src/extn/client/extn_client.rs
+++ b/core/sdk/src/extn/client/extn_client.rs
@@ -380,6 +380,7 @@ impl ExtnClient {
         };
         let new_context = { self.ripple_context.read().unwrap().clone() };
         let message = new_context.get_event_message();
+
         if propagate {
             trace!("Formed Context update event: {:?}", message);
             let c_message: CExtnMessage = message.clone().into();
@@ -390,7 +391,9 @@ impl ExtnClient {
                     trace!("Send to other client result: {:?}", send_res);
                 }
             }
-            Self::handle_vec_stream(message, self.event_processors.clone());
+            if self.has_event_listener(&message.target.as_clear_string()) {
+                Self::handle_vec_stream(message, self.event_processors.clone());
+            }
         } else {
             trace!("Context information is already updated. Hence not propagating");
         }
--- a/core/sdk/src/extn/client/extn_client.rs
+++ b/core/sdk/src/extn/client/extn_client.rs
@@ -294,16 +294,8 @@ impl ExtnClient {
                                 continue;
                             }
                         }
-                    } else {
-                        if is_context && self.has_event_listener(&message.target.as_clear_string())
-                        {
-                            Self::handle_vec_stream(message, self.event_processors.clone());
-                        } else {
-                            continue;
-                        }
                     }
                 }
-            } else {
                 let current_cap = self.sender.get_cap();
                 let target_contract = message.clone().target;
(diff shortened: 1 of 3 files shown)


core/sdk/src/extn/client/extn_client.rs
```diff
@@ -285,6 +285,14 @@ impl ExtnClient {
...
                                 ripple_context.deep_copy(context);
                             }
+                            let is_ripple_context =
+                                RippleContext::is_ripple_context(&message.payload).is_none();
+
+                            if self.has_event_listener(&message.target.as_clear_string()) {
+                                continue;
+                            } else if !is_ripple_context {
+                                warn!("No valid processors for the event {:?}", message);
+                            }
```
pahearn73, line 295: I'm not following the purpose of the `is_ripple_context` check, the ask in the ticket is unclear to me. @satlead can you clarify the ask there?

satlead: This was an incorrect warning and there shouldnt be one for Extensions which do not have a registered Ripple Context processor. This warning is intended for messages which is not ripple context and sent to an extension without a processor

pahearn73: Will we not already display a similar error when `handle_vec_stream()` is called? I'm trying to understand why it's necessary here.

pahearn73 (approved):

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -285,6 +285,14 @@ impl ExtnClient {
                                 let mut ripple_context = self.ripple_context.write().unwrap();
                                 ripple_context.deep_copy(context);
                             }
+                            let is_ripple_context =
+                                RippleContext::is_ripple_context(&message.payload).is_none();
+
+                            if self.has_event_listener(&message.target.as_clear_string()) {
```
satlead, line 291: Please add a negation here it should not proceed to send the ripple context if there are any listeners

nnaveen979: added the change please have a look

satlead (approved):

device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
```diff
@@ -1116,16 +1116,6 @@ impl ThunderDeviceInfoRequestProcessor {
 
         // If timezone or offset is None or empty
         if let Some(tz) = Self::get_timezone_and_offset(&state).await {
-            let cloned_state = state.clone();
```
satlead, line 1119: We get the Timezone and set it in Ripple context here. This sets up the initial context until the point the user changes the timezone. Are we removing this because it is adding that log?

pahearn73: Same question.

nnaveen979: @sdk, I removed it because i understood like, that cause's the warning log. I'll revert this change.

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #650, opened 2024-10-11
feat: Upgrade Telemetry and better logging levels

satlead:
## What

Adds better logging to help triage and measure telemetry

## Why

Telemetry currently offered in Ripple is very limiting, there is a need for measuring the time taken in each stage for every request.
Some of the Ripple logging have been too verbose for triaging and some production logs are lacking to help investigate issues faster.

## How

By addition of RpcStats object to RPCRequest during request processing and API message for Response dispatching
## Test

Tested by making calls to multiple Firebolt APIs and validating the logs

## Checklist

### DIFF
--- a/core/main/src/bootstrap/start_fbgateway_step.rs
+++ b/core/main/src/bootstrap/start_fbgateway_step.rs
@@ -15,6 +15,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use std::time::Instant;
+
 use crate::{
     firebolt::{
         firebolt_gateway::FireboltGateway,
@@ -40,7 +42,7 @@ use crate::{
     state::{bootstrap_state::BootstrapState, platform_state::PlatformState},
 };
 use jsonrpsee::core::{async_trait, server::rpc_module::Methods};
-use ripple_sdk::log::debug;
+use ripple_sdk::log::{debug, info};
 use ripple_sdk::{framework::bootstrap::Bootstep, utils::error::RippleError};
 pub struct FireboltGatewayStep;
 
@@ -108,6 +110,10 @@ impl Bootstep<BootstrapState> for FireboltGatewayStep {
             return Err(RippleError::BootstrapError);
         }
         TelemetryBuilder::send_ripple_telemetry(&state.platform_state);
+        info!(
+            "Ripple Total Bootstrap time: {}",
+            Instant::now().duration_since(state.start_time).as_millis()
+        );
         gateway.start().await;
 
         Err(RippleError::ServiceError)
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -19,7 +19,7 @@ use ripple_sdk::{
     api::{
         firebolt::fb_capabilities::JSON_RPC_STANDARD_ERROR_INVALID_PARAMS,
         gateway::rpc_gateway_api::{
-            ApiMessage, ApiProtocol, CallContext, JsonRpcApiResponse, RpcRequest,
+            ApiMessage, ApiProtocol, ApiStats, CallContext, JsonRpcApiResponse, RpcRequest,
         },
         session::AccountSession,
     },
@@ -46,7 +46,10 @@ use crate::{
     firebolt::firebolt_gateway::{FireboltGatewayCommand, JsonRpcError},
     service::extn::ripple_client::RippleClient,
     state::platform_state::PlatformState,
-    utils::router_utils::{return_api_message_for_transport, return_extn_response},
+    utils::router_utils::{
+        add_telemetry_status_code, get_rpc_header, return_api_message_for_transport,
+        return_extn_response,
+    },
 };
 
 use super::{
@@ -641,12 +644,11 @@ impl BrokerOutputForwarder {
                                             }
                                             response.id = Some(request_id);
 
-                                            let message = ApiMessage {
-                                                request_id: request_id.to_string(),
+                                            let message = ApiMessage::new(
                                                 protocol,
-                                                jsonrpc_msg: serde_json::to_string(&response)
-                                                    .unwrap(),
-                                            };
+                                                serde_json::to_string(&response).unwrap(),
+                                                request_id.to_string(),
+                                            );
 
                                             if let Some(session) = platform_state_c
                                                 .session_state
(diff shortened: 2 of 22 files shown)


core/main/src/bootstrap/start_fbgateway_step.rs
```diff
@@ -108,6 +110,10 @@ impl Bootstep<BootstrapState> for FireboltGatewayStep {
             return Err(RippleError::BootstrapError);
         }
         TelemetryBuilder::send_ripple_telemetry(&state.platform_state);
+        info!(
+            "Ripple Total Bootstrap time: {}",
+            Instant::now().duration_since(state.start_time).as_millis()
```
Vinodsathyaseelan, line 115: I assume that `info!()` log will be removed from production build. There is some performance impact on using `now() `. It is advised to use this function `sparingly`.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -641,12 +644,11 @@ impl BrokerOutputForwarder {
...
                                             response.id = Some(request_id);
 
-                                            let message = ApiMessage {
-                                                request_id: request_id.to_string(),
+                                            let message = ApiMessage::new(
                                                 protocol,
-                                                jsonrpc_msg: serde_json::to_string(&response)
-                                                    .unwrap(),
-                                            };
+                                                serde_json::to_string(&response).unwrap(),
```
Vinodsathyaseelan, line 649: unsafe unwrap here. Use pattern matching instead
`let response = match serde_json::to_string(&response) {...}`

core/main/src/broker/endpoint_broker.rs
```diff
@@ -696,13 +698,29 @@ impl BrokerOutputForwarder {
...
+                        let tm_str = get_rpc_header(&rpc_request);
                         // Step 2: Create the message
-                        let message = ApiMessage {
-                            request_id: request_id.to_string(),
-                            protocol: rpc_request.ctx.protocol.clone(),
-                            jsonrpc_msg: serde_json::to_string(&response).unwrap(),
-                        };
+                        let mut message = ApiMessage::new(
+                            rpc_request.ctx.protocol.clone(),
+                            serde_json::to_string(&response).unwrap(),
```
Vinodsathyaseelan, line 705: unsafe unwrap.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -696,13 +698,29 @@ impl BrokerOutputForwarder {
...
+                            request_id.to_string(),
+                        );
+                        let mut status_code: i64 = 1;
+                        if let Some(e) = &response.error {
+                            if let Some(Value::Number(n)) = e.get("code") {
+                                if let Some(v) = n.as_i64() {
+                                    status_code = v;
+                                }
+                            }
+                        }
```
Vinodsathyaseelan, line 715: The above code block can be replaced by

```
let status_code = response
    .error
    .as_ref()
    .and_then(|e| e.get("code"))
    .and_then(|v| v.as_i64())
    .unwrap_or(1);
```

Vinodsathyaseelan, line 715: The above code block can be replaced by 

```
let status_code = response
    .error
    .as_ref()
    .and_then(|e| e.get("code"))
    .and_then(|v| v.as_i64())
    .unwrap_or(1);
```

core/main/src/firebolt/firebolt_gateway.rs
```diff
@@ -312,33 +296,55 @@ impl FireboltGateway {
...
+    // in index 0
+    if let Ok(params) = serde_json::from_str::<Vec<serde_json::Value>>(&request.params_json) {
+        // Check if there are params
+        let param = params.get(1);
+        if param.is_none() {
+            // if params are necessary then handler or jq rule will fail
+            // This method can unblock other calls.
+            return Ok(());
+        }
+        let param = param.unwrap();
```
Vinodsathyaseelan, line 310: unsafe unwrap()

core/main/src/firebolt/firebolt_gateway.rs
```diff
@@ -312,33 +296,55 @@ impl FireboltGateway {
...
+        if let Some(rpc_method) = openrpc_validator.get_method_by_name(&method_name) {
+            // Get schema validator
+            let validator = openrpc_validator
+                .params_validator(major_version, &rpc_method.name)
+                .unwrap();
+            // validate
+            if let Err(errors) = validator.validate(param) {
+                let mut error_string = String::new();
+                for error in errors {
+                    error_string.push_str(&format!("{} ", error));
```
Vinodsathyaseelan, line 334: Any specific reasons to use `push_str` here instead of `write` ?

core/main/src/bootstrap/start_fbgateway_step.rs
satlead: This log is once per Ripple successful bootstrap.

core/main/src/broker/endpoint_broker.rs
satlead: Existing code and its Serialization of an existing rust object. Even the underlying Serde impl only returns OK

```
#[inline]
pub fn to_string<T>(value: &T) -> Result<String>
where
    T: ?Sized + Serialize,
{
    let vec = tri!(to_vec(value));
    let string = unsafe {
        // We do not emit invalid UTF-8.
        String::from_utf8_unchecked(vec)
    };
    Ok(string)
}
```

core/main/src/broker/endpoint_broker.rs
satlead: Existing code and its Serialization of an existing rust object. Even the underlying Serde impl only returns OK

```
#[inline]
pub fn to_string<T>(value: &T) -> Result<String>
where
    T: ?Sized + Serialize,
{
    let vec = tri!(to_vec(value));
    let string = unsafe {
        // We do not emit invalid UTF-8.
        String::from_utf8_unchecked(vec)
    };
    Ok(string)
}

File: core/main/src/firebolt/firebolt_gateway.rs (hunk 5 above)
Comment (satlead) [reply core/main/src/firebolt/firebolt_gateway.rs:310]: param is checked in line number 305 so its not unsafe

File: core/main/src/firebolt/firebolt_gateway.rs (hunk 6 above)
Comment (satlead) [reply core/main/src/firebolt/firebolt_gateway.rs:334]: This error string will not be printed here

**Review by Vinodsathyaseelan — APPROVED** [review]

File: core/main/src/state/openrpc_state.rs [hunk 7]
```diff
@@ -442,6 +446,28 @@ impl OpenRpcState {
             .unwrap()
             .extend(provider_relation_sets)
     }
+
```
pahearn73, line 449: Why not cache in the validator (eventually a crate) so it's universal?

core/main/src/broker/endpoint_broker.rs
Vinodsathyaseelan: Still it's a good practice to use the pattern matching code style. We have cleaned up many `unwrap()` usage with pattern matching.

core/main/src/state/openrpc_state.rs
satlead: Validator Crate uses serde to build the object it cannot have a thread safe box in it
#[derive(Debug, Serialize, Deserialize, Clone)]
pub struct FireboltOpenRpc {

core/main/src/broker/endpoint_broker.rs
satlead: Conflicted feel this is more readable

core/main/src/firebolt/rpc_router.rs
```diff
@@ -125,7 +130,27 @@ async fn resolve_route(
...
+        let request_id = req.ctx.request_id;
+        let status_code = if let Ok(r) = serde_json::from_str::<JsonRpcMessage>(&r) {
+            if let Some(ec) = r.error {
+                ec.code
+            } else {
+                1
+            }
+        } else {
+            1
+        };
```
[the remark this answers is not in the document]
satlead: Conflicted feel this is more readable

core/main/src/broker/endpoint_broker.rs
pahearn73: Either way is fine.

pahearn73 (approved):

PR #655, opened 2024-10-14
feat: Implement UserDataMigrator Framework - Rule based Shim Layer

Vinodsathyaseelan:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/rules_engine.rs
+++ b/core/main/src/broker/rules_engine.rs
@@ -89,15 +89,21 @@ pub struct Rule {
     // Not every rule needs transform
     #[serde(default)]
     pub transform: RuleTransform,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub filter: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub endpoint: Option<String>,
 }
 
 #[derive(Debug, Clone, Deserialize, Default, Serialize)]
 pub struct RuleTransform {
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub request: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub response: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub event: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub event_decorator_method: Option<String>,
 }
 
--- a/core/main/src/broker/rules_engine.rs
+++ b/core/main/src/broker/rules_engine.rs
@@ -209,10 +209,7 @@ impl RuleEngine {
         None
     }
     pub fn get_rule_by_method(&self, method: &str) -> Option<Rule> {
-        if let Some(rule) = self.rules.rules.get(&method.to_lowercase()).cloned() {
-            return Some(rule);
-        }
-        None
+        self.rules.rules.get(&method.to_lowercase()).cloned()
     }
 }
 
--- a/core/main/src/broker/rules_engine.rs
+++ b/core/main/src/broker/rules_engine.rs
@@ -24,7 +24,7 @@ use ripple_sdk::{
     serde_json::Value,
     utils::error::RippleError,
 };
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::{fs, path::Path};
 
@@ -83,7 +83,7 @@ pub enum RuleEndpointProtocol {
     Thunder,
 }
 
-#[derive(Debug, Clone, Deserialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Rule {
     pub alias: String,
     // Not every rule needs transform
@@ -93,7 +93,7 @@ pub struct Rule {
     pub endpoint: Option<String>,
 }
 
-#[derive(Debug, Clone, Deserialize, Default)]
+#[derive(Debug, Clone, Deserialize, Default, Serialize)]
 pub struct RuleTransform {
     pub request: Option<String>,
     pub response: Option<String>,
@@ -199,6 +199,12 @@ impl RuleEngine {
                 "Rule not available for {}, hence falling back to extension handler",
                 rpc_request.method
             );
+        }
+        None
+    }
+    pub fn get_rule_by_method(&self, method: &str) -> Option<Rule> {
+        if let Some(rule) = self.rules.rules.get(&method.to_lowercase()).cloned() {
+            return Some(rule);
         }
         None
     }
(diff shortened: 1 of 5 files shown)


core/main/src/broker/rules_engine.rs
```diff
@@ -202,6 +208,12 @@ impl RuleEngine {
         }
         None
     }
+    pub fn get_rule_by_method(&self, method: &str) -> Option<Rule> {
+        if let Some(rule) = self.rules.rules.get(&method.to_lowercase()).cloned() {
+            return Some(rule);
+        }
+        None
+    }
```
pahearn73, line 216: Nitpick: This could just be:

`
pub fn get_rule_by_method(&self, method: &str) -> Option<Rule> {
    self.rules.rules.get(&method.to_lowercase()).cloned()
}
`

core/main/src/broker/thunder/user_data_migrator.rs
```diff
@@ -0,0 +1,918 @@
...
+use tokio_tungstenite::{tungstenite::Message, WebSocketStream};
+
+// TBD get the storage dir from manifest or other Ripple config file
+const RIPPLE_STORAGE_DIR: &str = "/opt/persistent/ripple";
+const USER_DATA_MIGRATION_CONFIG_FILE_NAME: &str = "user_data_migration_config.json";
+
+#[derive(Debug)]
+enum UserDataMigratorError {
+    ThunderRequestError(String),
+    ResponseError(String),
```
pahearn73, line 62: Is `ResponseError` a thunder response error and if so should it follow the naming convention for request errors, e.g. `ThunderResponseError`?

core/main/src/broker/thunder_broker.rs
```diff
@@ -105,14 +156,43 @@ impl ThunderBroker {
...
+                                        let _feed = ws_tx.feed(tokio_tungstenite::tungstenite::Message::Text(r)).await;
+                                        let _flush = ws_tx.flush().await;
+                                    }
+                                }
+                                else {
+                                    // empty request means plugin is activated and ready to process the request
+                                    // Intercept the request for data migration
+                                    let mut request_consumed = false;
+                                    if let Some(user_data_migrator) = broker_c.data_migrator.clone() {
+                                        request_consumed = user_data_migrator.intercept_broker_request(&broker_c, ws_tx_wrap.clone(), &mut request).await;
```
brendanobra, line 176: What does the failure case  produce, and is it recoverable (maybe I'm asking questions that are not "now problems")

core/main/src/broker/thunder_broker.rs
```diff
@@ -168,6 +249,13 @@ impl ThunderBroker {
         new_response
     }
 
+    fn get_id_from_result(result: &[u8]) -> Option<u64> {
+        if let Ok(data) = serde_json::from_slice::<JsonRpcApiResponse>(result) {
```
brendanobra, line 253: nitpick: can this be a map() or similar?

core/main/src/broker/thunder/user_data_migrator.rs
```diff
@@ -0,0 +1,918 @@
...
+            );
+            broker.unregister_custom_callback(request_id).await;
+            return Err(e);
+        }
+
+        // get the response from the custom callback
+        Self::wait_for_response(self.response_rx.clone(), broker.clone(), request_id).await
+    }
+
+    fn transform_requets_params(
```
pahearn73, line 461: Typo: `transform_requets_params`

core/main/src/broker/rules_engine.rs
Vinodsathyaseelan: Agreed. Will change as suggested.

core/main/src/broker/thunder/user_data_migrator.rs
Vinodsathyaseelan: Will change this to `ThunderResponseError` for making it consistent.

core/main/src/broker/thunder/user_data_migrator.rs
```diff
@@ -0,0 +1,918 @@
...
+    migrated: bool,
+}
+
+type MigrationMap = HashMap<String, MigrationConfigEntry>;
+#[derive(Clone, Debug)]
+pub struct UserDataMigrator {
+    migration_config: Arc<Mutex<MigrationMap>>, // persistent migration map
+    config_file_path: String,                   // path to the migration map file
+    response_tx: Sender<BrokerOutput>,
+    response_rx: Arc<Mutex<Receiver<BrokerOutput>>>,
```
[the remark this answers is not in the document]
Vinodsathyaseelan: This is internal to DataMigrator and will be shared between read/write functions of multiple plugins calls in async operation. Adding a spawned receiver loop will make the operation more of synchronous and I was running into a deadlock scenario. May be a candidate for next iteration.

core/main/src/broker/thunder/user_data_migrator.rs
```diff
@@ -0,0 +1,918 @@
...
+
+        info!(
+            "intercept_broker_request: No migration entry found for method: {:?}",
+            method
+        );
+        // Continue with the original request if no migration entry is found
+        false
+    }
+
+    async fn handle_setter_request(
```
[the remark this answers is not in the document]
Vinodsathyaseelan: `handle_setter_request` will be called from `intercept_broker_request`  for those firebolt set api calls listed in migration config file. 

The use case here is to always update `LegacyStorage` with the **new value**  while continuing with the normal set operation of new plugin. The `LegacyStorage` update is required to avoid data loss in case of rollback.

core/main/src/broker/thunder_broker.rs
Vinodsathyaseelan: If the `thunder_broker` has `user_data_migrator` installed, all requests will be intercepted for migration processing. 

`request_consumed` will be false for the following scenarios.

1) The Requested FireBolt API is not listed in the migration configuration file 
2) There are errors in the migration configuration entries.
3) Migration is not required, for example, if there are no entries in Legacy Storage.
4) The request is partially consumed. For example,  in the case of a setter Firebolt API, the migrator will set the new value in Legacy Storage  and allow the `thunder_broker` to continue the rest of the operation of updating the value in the new plugin.

core/main/src/broker/thunder_broker.rs
Vinodsathyaseelan: Changed to

```
        serde_json::from_slice::<JsonRpcApiResponse>(result)
            .ok()
            .and_then(|data| data.id)
```

brendanobra (approved):

pahearn73 (approved):

PR #673, opened 2024-11-04
feat: RPPL-2681 thunderBroker migration workaround changes

nnaveen979:
## What

thunderBroker migration changes added.

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/client/thunder_async_client.rs
+++ b/device/thunder_ripple_sdk/src/client/thunder_async_client.rs
@@ -1,0 +1,304 @@
+// Copyright 2023 Comcast Cable Communications Management, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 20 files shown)


device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,304 @@
...
+                    "Broker with {} failed with retry for last {} secs in {}",
+                    endpoint, index, port
+                );
+            }
+            index += 1;
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+    }
+
+    fn prepare_request(&self, request: &ThunderAsyncRequest) -> Result<Vec<String>, RippleError> {
```
Vinodsathyaseelan, line 122: Do you have any documentation related to the new changes ?  Is ThunderBroker a singleton object ?  If not, I have concern about multiple instances of ThunderBroker.  In that case you may need to separate state management instance out of Thunder Broker. 

I have also cleaned up  `prepare_request` in the original implementation.  Thunder Plugin state checking and activation have been moved to `check_and_generate_plugin_activation_request`  helper function,

nnaveen979:
not needed anymore, closing this PR.

PR #674, opened 2024-11-05
RPPL-2658 Cleanup handlers for Localization and secondscreen

sakshihcst:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/thunder_broker.rs
+++ b/core/main/src/broker/thunder_broker.rs
@@ -486,8 +486,8 @@ mod tests {
         }
     }
 
-    #[tokio::test]
     #[ignore]
+    #[tokio::test]
     async fn test_thunderbroker_start() {
         let (tx, mut _rx) = mpsc::channel(1);
         let (sender, mut rec) = mpsc::channel(1);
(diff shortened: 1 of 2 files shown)


core/main/src/broker/broker_utils.rs
```diff
@@ -111,4 +112,68 @@ impl BrokerUtils {
             data: None,
         }))
     }
+
+    pub async fn get_language(ctx: &CallContext, state: &PlatformState) -> RpcResult<String> {
```
Vinodsathyaseelan, line 116: Make this function generic by accepting `method` name as an input param. After that,  change the usage in caller by passing `localization.language` or `localization.countryCode` as param to the new function.  Remove `get_language` and `get_country_code` functions from BrokerUtils

brendanobra (approved):

core/main/src/state/metrics_state.rs
```diff
@@ -320,9 +330,11 @@ impl MetricsState {
 
         let coam = Self::get_persistent_store_bool(state, PERSISTENT_STORAGE_KEY_COAM).await;
 
-        let country = StorageManager::get_string(state, StorageProperty::CountryCode)
+        let country = BrokerUtils::process_internal_main_request(state, "localization.countryCode")
             .await
-            .ok();
+            .ok()
+            .and_then(|val| from_value::<String>(val).ok());
```
brendanobra, line 336: I think this will set `country` to `true` or `false`

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #675, opened 2024-11-07
feat: thunderBroker migration changes with config switch

nnaveen979:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -25,21 +25,81 @@ use ripple_sdk::{
...
+                pool_size: None,
+                plugin_param: None,
+                thunder_connection_state: None,
+            };
+
+            let thndr_boot_stateclient = ThunderBootstrapStateWithClient {
+                prev: thndr_boot_statecfg,
+                state: thunder_state,
+            };
+            return Some(thndr_boot_stateclient);
```
satlead, line 88: You need to call SetupThunderProcessor::setup(state.clone()).await;

device/thunder_ripple_sdk/src/client/thunder_client2.rs
satlead, line 1: Can we delete this file?

device/thunder_ripple_sdk/src/client/thunder_client2.rs
```diff
@@ -0,0 +1,157 @@
...
+        tokio::spawn(async move {
+            loop {
+                request_tr = client_c.client.start("", request_tr).await;
+                error!("Thunder disconnected so reconnecting")
+            }
+        });
+
+        tokio::spawn(async move {
+            while let Some(v) = response_tr.recv().await {
+                // check with thunder client callbacks and subscriptions
```
pahearn73, line 137: @satlead : What's the purpose of `ThunderClientManager` here, is it supposed to receive requests and send them to a new thunder broker insteance, and then it also listens for responses from that broker and sends the response back to the requestor? Trying to understand what @nnaveen979 needs to do here.

satlead: Thats a sample code he is supposed to delete this file

core/sdk/src/utils/rpc_utils.rs
```diff
@@ -20,3 +20,13 @@ use jsonrpsee::core::Error;
 pub fn rpc_err(msg: impl Into<String>) -> Error {
     Error::Custom(msg.into())
 }
+
+pub fn extract_tcp_port(url: &str) -> String {
+    let url_split: Vec<&str> = url.split("://").collect();
+    if let Some(domain) = url_split.get(1) {
+        let domain_split: Vec<&str> = domain.split('/').collect();
+        domain_split.first().unwrap().to_string()
```
brendanobra, line 28: What happens when this unwrap fails? Consider using a result type (Result<String, Err>) or , if it is possible, provide a default.

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -21,25 +21,105 @@ use crate::{
...
+
+fn gateway_default() -> String {
+    String::from(GATEWAY_DEFAULT)
+}
 
 pub async fn boot_thunder(
-    state: ExtnClient,
+    ext_client: ExtnClient,
     plugin_param: ThunderPluginBootParam,
 ) -> Option<ThunderBootstrapStateWithClient> {
```
brendanobra, line 53: Consider refactoring with Result<_> return type - what does the caller do when this returns None?

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -21,25 +21,105 @@ use crate::{
...
-            error!("Unable to connect to Thunder, error in ThunderPoolStep");
+    //by default enabling the thunderBroker
+    if ext_client.get_bool_config("use_with_thunder_broker") {
+        info!("Using thunder broker");
+        let mut extn_client = ext_client.clone();
+        let mut gateway_url = url::Url::parse(GATEWAY_DEFAULT).unwrap();
+        let extn_message_response: Result<ExtnMessage, RippleError> =
+            extn_client.request(Config::PlatformParameters).await;
+
+        if let Ok(message) = extn_message_response {
```
brendanobra, line 63: .map()ing maybe be more readable/maintainable.

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
```diff
@@ -98,10 +103,10 @@ impl ThunderPoolStep {
         }
 
         let client = ThunderClientPool::start(
-            url,
+            url.clone().unwrap(),
```
brendanobra, line 106: Unwrap

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
```diff
@@ -98,10 +103,10 @@ impl ThunderPoolStep {
...
             Some(plugin_manager_tx),
             thunder_connection_state.clone(),
-            pool_size - 1,
+            pool_size.unwrap() - 1,
```
brendanobra, line 109: unwrap?

core/sdk/src/api/device/device_operator.rs
```diff
@@ -44,6 +44,33 @@ pub enum DeviceChannelRequest {
     Unsubscribe(DeviceUnsubscribeRequest),
 }
 
+impl DeviceChannelRequest {
+    pub fn get_callsign_method(&self) -> (String, String) {
```
pahearn73, line 48: This is thunder-specific so it should probably go into `thunder_ripple_sdk` and called only by thunder-specific code.

core/sdk/src/utils/rpc_utils.rs
pahearn73: Agree, please return `Result` and check during use so we don't enter endless loops when creating websockets.

device/thunder_ripple_sdk/Cargo.toml
```diff
@@ -56,8 +59,9 @@ csv = "1.1"  # Allowing minor updates
 home = { version = "=0.5.5", optional = true }
 tree_magic_mini = { version = "=3.0.3", optional = true }
 rstest = { version = "0.18.2", optional = true, default-features = false }
+tokio-tungstenite = { workspace = true, features = ["handshake"] }
+futures-util = { version = "0.3.28", features = ["sink", "std"], default-features = false}
```
pahearn73, line 63: Can you please confirm we need all of these new dependencies above?

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -21,25 +21,105 @@ use crate::{
...
-        if let Ok(state) = ThunderPoolStep::setup(state).await {
-            SetupThunderProcessor::setup(state.clone()).await;
-            return Some(state);
-        } else {
-            error!("Unable to connect to Thunder, error in ThunderPoolStep");
+    //by default enabling the thunderBroker
+    if ext_client.get_bool_config("use_with_thunder_broker") {
+        info!("Using thunder broker");
+        let mut extn_client = ext_client.clone();
+        let mut gateway_url = url::Url::parse(GATEWAY_DEFAULT).unwrap();
```
pahearn73, line 59: Is this always going to use the default even if there is a value in the manifest? I think you want to pull from config and let the config logic decide whether or not to give you the default.

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -21,25 +21,105 @@ use crate::{
...
+                    );
+                }
+                if let Ok(host_override) = std::env::var("DEVICE_HOST") {
+                    gateway_url.set_host(Some(&host_override)).ok();
+                }
+            }
         }
+
+        if let Ok(thndr_client) =
+            ThunderClientBuilder::get_client(Some(gateway_url), None, None, None, None, true).await
```
pahearn73, line 89: `ThunderClientBuilder::get_client` shouldn't take `url: Option<Url>` as a parameter, just `url: Url`, since it's a required parameter whether or not ThunderBroker is enabled. Fixing that will also get rid of unwanted unwraps later.

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -21,25 +21,105 @@ use crate::{
...
+            if let Ok(state) = ThunderPoolStep::setup(state).await {
+                SetupThunderProcessor::setup(state.clone()).await;
+                return Some(state);
+            } else {
+                error!("Unable to connect to Thunder, error in ThunderPoolStep");
+            }
+        } else {
+            error!("Unable to connect to Thunder, error in ThunderGetConfigStep");
+        }
+        None
```
pahearn73, line 123: Whether or not `use_with_thunder_broker` is set, both of these condition blocks do something like:

```
SetupThunderProcessor::setup(state).await;
return Some(state); // (or None)
```
...I would just refactor to clean it up a little, e.g.

```
let state = if use_with_thunder_broker {...} else {...};

if let Some(s) = state {
    SetupThunderProcessor::setup(state).await;
}

state
```

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
```diff
@@ -43,13 +43,18 @@ impl ThunderPoolStep {
         let pool_size = state.pool_size;
         let url = state.url.clone();
         let thunder_connection_state = state.thunder_connection_state.clone();
-        if pool_size < 2 {
+        if pool_size < Some(2) {
```
pahearn73, line 46: We should first check to see if `pool_size.is_some()` and error/exit if so, then you also don't have to do an `unwrap` later, e.g.:

```
let pool_size = match state.pool_size {
    Some(s) => s,
    None => {
        error!("....");
        return Err(RippleError::BootstrapError);
    }
};
````

File: device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs [hunk 14]
```diff
@@ -43,13 +43,18 @@ impl ThunderPoolStep {
...
             warn!("Pool size of 1 is not recommended, there will be no dedicated connection for Controller events");
             return Err(RippleError::BootstrapError);
         }
         let controller_pool = ripple_sdk::tokio::time::timeout(
             Duration::from_secs(10),
-            ThunderClientPool::start(url.clone(), None, thunder_connection_state.clone(), 1),
+            ThunderClientPool::start(
+                url.clone().unwrap(),
```
pahearn73, line 53: I don't think `url` needs to be an `Option`.

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
```diff
@@ -71,7 +76,7 @@ impl ThunderPoolStep {
         let expected_plugins = state.plugin_param.clone();
         let tc = Box::new(controller_pool);
         let (plugin_manager_tx, failed_plugins) =
-            PluginManager::start(tc, expected_plugins.clone()).await;
+            PluginManager::start(tc, expected_plugins.clone().unwrap()).await;
```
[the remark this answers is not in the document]
pahearn73: Yeah does `ThunderBootstrapStateWithConfig.plugin_param` need to be an `Option`?

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+        Self {
+            id: get_next_id(),
+            request,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ThunderAsyncResponse {
+    pub id: Option<u64>,
```
pahearn73, line 62: Why is `id` an `Option`? Looks like you supply it for errors but not for regular responses, why not just always include it? Seems like it's always available...

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+    pub result: Result<JsonRpcApiResponse, RippleError>,
+}
+
+impl ThunderAsyncResponse {
+    fn new_response(response: JsonRpcApiResponse) -> Self {
+        Self {
+            id: None,
```
pahearn73, line 69: Per above

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+            result: Ok(response),
+        }
+    }
+
+    fn new_error(id: u64, e: RippleError) -> Self {
+        Self {
+            id: Some(id),
```
pahearn73, line 76: Per above

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+                    endpoint, index, port
+                );
+            }
+            index += 1;
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+    }
+
+    fn prepare_request(&self, request: &ThunderAsyncRequest) -> Result<Vec<String>, RippleError> {
+        println!("@@@NNA-----> in start prepare request id :{}", request.id);
```
pahearn73, line 172: Reminder: Remove debugging `println!`s

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+    }
+
+    fn new_error(id: u64, e: RippleError) -> Self {
+        Self {
+            id: Some(id),
+            result: Err(e),
+        }
+    }
+
+    pub fn get_event(&self) -> Option<String> {
```
pahearn73, line 81: Should this be called `get_method` since it's returning the method name and not an event per se?

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+        if let Ok(e) = &self.result {
+            return e.method.clone();
+        }
+        None
+    }
+
+    pub fn get_id(&self) -> Option<u64> {
+        match &self.result {
+            Ok(response) => response.id,
```
pahearn73, line 90: I'm confused by this, why doesn't it just return `Self::id`? This goes back to the question of why id is an `Option`.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+            Err(_) => None,
+        }
+    }
+
+    pub fn get_device_resp_msg(&self, sub_id: Option<String>) -> Option<DeviceResponseMessage> {
+        println!("@@NNA...get_device_resp_msg res:{:?}", self.result);
```
pahearn73, line 96: Remove debug logging

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+        let json_resp = match &self.result {
+            Ok(json_resp_res) => {
+                println!("@@NNA...get_device_resp_msg resp::{:?}", json_resp_res);
+                json_resp_res
+            }
+            _ => return None,
+        };
+
+        let mut device_response_msg = None;
```
pahearn73, line 105: The below conditional blocks are the only places where `DeviceResponseMessage::new` is called, it would be cleaner and more encapsulated to just do something like:

`let device_response_msg = DeviceResponseMessage::create(json_resp, sub_id);`

...where `DeviceResponseMessage::create()` returns `Option<DeviceResponseMessage>`. Then do all that conditional stuff to create the DeviceResponseMessage inside the new `create()` method.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+        serde_json::from_slice::<JsonRpcApiResponse>(result)
+            .ok()
+            .and_then(|data| data.id)
+    }
+
+    pub fn get_sender(&self) -> BrokerSender {
+        self.sender.clone()
+    }
+
+    pub fn get_thndrasync_callback(&self) -> BrokerCallback {
```
pahearn73, line 135: Rename this to `get_callback()`.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+        self.callback.clone()
+    }
+    async fn create_ws(
+        endpoint: &str,
+    ) -> (
+        SplitSink<WebSocketStream<TcpStream>, Message>,
+        SplitStream<WebSocketStream<TcpStream>>,
+    ) {
+        info!("Broker Endpoint url {}", endpoint);
```
pahearn73, line 144: this should probably be `debug!()`

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+
+        let url = url::Url::parse(endpoint).unwrap();
```
pahearn73, line 146: Get rid of `url` since it's only used for logging and you're doing an unwrap because of it. Just log `endpoint` it's already a `&str`.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+            Some(v) => {
+                println!(
+                    "@@@NNA....status_manager.get_status returned some value. v:{:?} ",
+                    v
+                );
+                v.clone()
+            }
+            None => {
+                println!("@@@NNA....status_manager.get_status return none.");
+                self.status_manager
```
pahearn73, line 194: I see that `ThunderAsyncClient` has it's own plug-in status manager, this doesn't seem right since we already have one in `thunder_plugins_status_mgr.rs`. Does the existing manager not provide something you need? If not, then instead of creating another we should look into refactoring the existing `StatusManager` so that it suits both new and existing needs. Let's discuss what is needed to do this.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+                        Err(e) => {
+                            error!("Broker Websocket error on read {:?}", e);
+                            // Time to reconnect Thunder with existing subscription
+                            break;
+                        }
+                    }
+                },
+                Some(request) = tr.recv() => {
+                    debug!("Got request from receiver for broker {:?}", request);
+                    debug!("Got request from receiver for broker for method{:?}", request.request.get_callsign_method());
```
pahearn73, line 358: Redundant `debug!` call.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,617 @@
...
+                    match client_c.prepare_request(&request) {
```
pahearn73, line 359: What does`prepare_request()` actually return? a `Vec<String>` but what are they? It looks like we're routing messages here but it's not clear to me what the messages are, where they're from or where they're going, maybe we can clarify that with clearer variable names and comments?

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -22,6 +22,8 @@ use std::sync::Arc;
 use jsonrpsee::core::client::{Client, ClientT, SubscriptionClientT};
 use jsonrpsee::ws_client::WsClientBuilder;
 
+use crate::thunder_state::ThunderConnectionState;
```
pahearn73, line 25: // <pca> YAH </pca>

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -51,18 +53,108 @@ use ripple_sdk::{
...
 use std::{env, process::Command};
 
+pub type BrokerSubMap = HashMap<String, DeviceResponseSubscription>;
+pub type BrokerCallbackMap = HashMap<u64, Option<OneShotSender<DeviceResponseMessage>>>;
+
+#[derive(Debug)]
+pub struct ThunderClientManager;
+
+impl ThunderClientManager {
+    fn manage(
```
pahearn73, line 78: Maybe rename this to `start()` to align with what you named the same for `ThunderAsyncClient`.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -191,20 +287,67 @@ impl ThunderClient {
...
+            );
+            self.send_message(message).await;
 
-        match rx.await {
-            Ok(response) => response,
-            Err(_) => DeviceResponseMessage {
-                message: Value::Null,
-                sub_id: None,
-            },
+            // match rx.await {
```
pahearn73, line 303: Remove.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -191,20 +287,67 @@ impl ThunderClient {
...
+                Ok(response) => {
+                    println!("@@@NNA....deviceoperator response received:{:?}", response);
+                    response
+                }
+                Err(_) => DeviceResponseMessage {
+                    message: Value::Null,
+                    sub_id: None,
+                },
+            }
+            // DeviceResponseMessage {
```
pahearn73, line 347: Remove.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -542,120 +793,173 @@ impl ThunderClientBuilder {
...
-            plugin_manager_tx: pmtx_c,
-            subscriptions: Some(subscriptions),
-        })
+            Ok(ThunderClient {
+                sender: Some(s),
+                pooled_sender: None,
+                id: uid,
+                plugin_manager_tx: pmtx_c,
+                subscriptions: Some(subscriptions),
+                thndr_asynclient: None,
```
pahearn73, line 931: Can we rename and spell it out e.g. `thunder_async_client`?

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -51,18 +53,108 @@ use ripple_sdk::{
...
+        tokio::spawn(async move {
+            println!("@@@NNA----> in thunder::manage response block entrance...");
+            while let Some(response) = response_tr.recv().await {
+                println!(
+                    "@@@NNA...manage fn response block..with response{:?}",
+                    response
+                );
+                if let Some(id) = response.get_id() {
+                    println!("@@@NNA...manage fn response id: {:?}", id);
+                    if let Some(cb) = client.clone().broker_callbacks {
```
pahearn73, line 105: `cb` doesn't hold a single callback it is a map of callbacks, so rename to `broker_callbacks`.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -51,18 +53,108 @@ use ripple_sdk::{
...
+                        println!(
+                            "@@@NNA...manage fn broker_callbacks check inside..callback:{:?}",
+                            cb
+                        );
+                        let mut callback = cb.write().unwrap();
```
pahearn73, line 110: As above.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -51,18 +53,108 @@ use ripple_sdk::{
...
+                        println!(
+                            "@@@NNA.... after cb.write fn unwrap callback: {:?}",
+                            callback
+                        );
+                        if let Some(Some(c)) = callback.remove(&id) {
```
pahearn73, line 115: As above, and then `c` can be more clearly renamed to `callback`.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -213,28 +356,87 @@ impl DeviceOperator for ThunderClient {
...
+                        response
+                    );
+                    response
+                }
+                Err(_) => DeviceResponseMessage {
+                    message: Value::Null,
+                    sub_id: None,
+                },
+            }
+        } else if let Some(subscribe_request) = self.check_sub(&request, handler.clone()) {
```
pahearn73, line 390: Maybe we can rename `check_sub()` to better describe what it does, it's not clear what's happening here.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -213,28 +356,87 @@ impl DeviceOperator for ThunderClient {
...
+            let (tx, rx) = oneshot::channel::<DeviceResponseMessage>();
+            self.add_to_callback(&subscribe_request, tx);
```
pahearn73, line 392: Shouldn't `add_to_callback()` be named `add_callback()`? Looks like we're adding a callback to a map not adding anything to a callback.

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -213,28 +356,87 @@ impl DeviceOperator for ThunderClient {
...
+                        response
+                    );
+                    response
+                }
+                Err(_) => DeviceResponseMessage {
+                    message: Value::Null,
+                    sub_id: None,
+                },
+            }
+            //rx.await.unwrap()
```
pahearn73, line 415: Remove

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -213,28 +356,87 @@ impl DeviceOperator for ThunderClient {
...
+        if !self.use_thunderbroker {
+            let message = ThunderUnsubscribeMessage {
+                module: request.module,
+                event_name: request.event_name,
+                subscription_id: None,
+            };
+            let msg = ThunderMessage::ThunderUnsubscribeMessage(message);
+            self.send_message(msg).await;
+        } else {
+            // unsubscribe() deprecate
```
pahearn73, line 438: Shouldn't we be handling this scenario here for ThunderBroker?

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -476,6 +678,55 @@ impl ThunderClient {
...
+                sub_id: request.clone().sub_id,
+                handlers: vec![handler],
+            };
+
+            broker_subscriptions.insert(key, dev_resp_sub);
+            Some(async_request)
+        }
+    }
+
+    // if only one handler cleanup
```
pahearn73, line 726: Remove

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -542,120 +793,173 @@ impl ThunderClientBuilder {
     }
 
     pub async fn get_client(
```
pahearn73, line 795: `get_client` IMO has always be poorly named, it does a lot of stuff and now even more depending on if ThunderBroker is used or not. I feel like we should rename it and also consider moving the spawned receiver loop out of this function and into it's own. This is all really just part of start-up and getting things set up and not just "getting a client".

device/thunder_ripple_sdk/src/utils.rs
```diff
@@ -109,3 +112,10 @@ pub fn get_error_value(error: &Error) -> Value {
     }
     Value::Null
 }
+
+static ATOMIC_ID: AtomicU64 = AtomicU64::new(0);
+
+pub fn get_next_id() -> u64 {
```
pahearn73, line 118: Looks like `EndpointBrokerState` already does this, I suggest moving this out of here and into somewhere like `Ripple/core/sdk/src/utils`. Then refactor `EndpointBrokerState` to use the same.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: for events in response id comes as None. for this reason here id is Option.

for eg: in below mentioned log is response msg for PowerState change event.
log:
get_device_resp_msg res:Ok(JsonRpcApiResponse { jsonrpc: "2.0", id: None, result: None, error: None, method: Some("client.events.onSystemPowerStateChanged"), params: Some(Object {"currentPowerState": String("ON"), "powerState": String("LIGHT_SLEEP")}) })

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: cleaned up.

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: cleaned up

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: naming changed

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: check_sub() renamed to add_subscription_handler() and added more comments to describe what actually happens inside function

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: added more description in comments inside function definition.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: removed the extra debug log

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: changed the name

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: removed all print logs added by me

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: for event responses we don't get the id in response for that reason making the id field as optional in ThunderAsyncResponse

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: as above

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: as above

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
nnaveen979: changed now.

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
nnaveen979: done

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
nnaveen979: i tried to do the same way as suggested above but its throwing mismatch error

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
nnaveen979: done

core/sdk/src/utils/rpc_utils.rs
nnaveen979: changed now

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
nnaveen979: done

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
nnaveen979: done

core/sdk/src/api/device/device_operator.rs
nnaveen979: moved to thunder_async_client.rs file

device/thunder_ripple_sdk/Cargo.toml
nnaveen979: these are auto generated, but i found without these dependencies cargo build is failed. so these dependencies are needed.

device/thunder_ripple_sdk/src/utils.rs
nnaveen979: done

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
nnaveen979: this code is untouched. boot_thunder() invoked from 2 places which have different handling. 
ripple-eos-thunder-extn/src/bootstrap/boot_thunder_channel.rs here fn boot_thunder_channel() have handling for Option return type. but the other place Ripple/device/thunder/src/bootstrap/boot_thunder_channel.rs have nothing to do when there is none return type comes.

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: renamed and as we discussed moved code into separate function for not using thunderbroker.

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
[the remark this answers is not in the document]
nnaveen979: as of now only fn boot_thunder in boot_thunder.rs we use ThunderBootstrapStateWithConfig.

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
Vinodsathyaseelan: @nnaveen979 : I have refactored this code in thunder_broker.rs https://github.com/rdkcentral/Ripple/blob/main/core/main/src/broker/thunder_broker.rs#L164 

Some of the plugin activation check logic included in prepare_request() has been moved out of that function and  added a new function. Refer the newly added function `check_and_generate_plugin_activation_request ` 

Btw, what is the decision behind duplicating the thunder_broker code in thunder_async_client.rs ? Can we re-use the existing code ?

Vinodsathyaseelan (changes requested):
.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: thanks for the update @Vinodsathyaseelan , i'll look into those changes. in thunder_async_client we use tokio_tungstenite (async run time) which is new compared to thunder_broker. in this new implementation there is no thunder_client pooling.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -204,7 +201,7 @@ impl Default for BrokerCallback {
     }
 }
 
-static ATOMIC_ID: AtomicU64 = AtomicU64::new(0);
+// static ATOMIC_ID: AtomicU64 = AtomicU64::new(0);
```
pahearn73, line 204: Wasn't this being moved to the SDK only because we were talking about making the plug-in status manager handle both BrokerRequest and ThunderAsyncRequest types (but we are no longer doing that)?

core/sdk/src/api/device/device_operator.rs
pahearn73: Did this get moved? Seems like it's still here.

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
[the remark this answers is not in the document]
pahearn73: Ok we can run with that for now. Can you address @brendanobra 's unwrap concern though? Get rid of unwrap, if `expected_plugins.is_none()` then make `ThunderPoolStep::setup` return an error.

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
```diff
@@ -80,7 +83,7 @@ impl ThunderPoolStep {
             );
             loop {
                 let failed_plugins = PluginManager::activate_mandatory_plugins(
-                    expected_plugins.clone(),
+                    expected_plugins.as_ref().unwrap().clone(),
```
pahearn73, line 86: Same with unwrap here.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
pahearn73: Should `id` here be `response.id` and not `None`?

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
pahearn73: `ThunderAsyncResponse` isn't necessarily always an event though is it? Like it could be a response from a regular thunder request right?

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,564 @@
...
+                },
+                Some(request) = tr.recv() => {
+                    debug!("Got request from receiver for broker {:?}", request);
+                    // here prepare_request will check the plugin status and add json rpc format
+                    match client_c.prepare_request(&request) {
+                        Ok(updated_request) => {
+                            debug!("Sending request to broker {:?}", updated_request);
+                            for r in updated_request {
+                                let url_clone = url.to_string();
+                                tokio::spawn({
```
pahearn73, line 292: The thread spawned here should ONLY:

1. Create a new websocket connection to thunder.
2. Send the thunder request.
3. Await the response.
4. Call `Self::handle_jsonrpc_response()`

...it should not be looping on a `tokio::select!()` or checking the status manager for anything because that's being done on the socket you initially created when `start()` is called.

pahearn73: Also, I suggest putting the steps I described above into a new function/method e.g. 

`tokio::spawn({process_request(r);});`

...for readability.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,564 @@
...
+                                                                }
+                                                                else {
+                                                                    //let _id = Self::get_id_from_result(t.as_bytes()); for debug purpose
+                                                                    // send the incoming text without context back to the sender
+                                                                    Self::handle_jsonrpc_response(t.as_bytes(),callback.clone()).await
+                                                                }
+                                                            }
+                                                        },
+                                                        Err(e) => {
+                                                            error!("Broker Websocket error on read {:?}", e);
```
pahearn73, line 321: Since `thunder_async_client.rs` was based off of `thunder_broker.rs`, there are a lot of logs, objects, and variables in this file that reference `broker`. Can you run through this file and do a general clean-up of logs and names?

core/sdk/src/api/device/device_operator.rs
nnaveen979: i tried to move the whole struct and related implementations from device_operator.rs to thunder_async_client.rs file  but unfortunately black_Duck check fails continuously. when i revert that change then black_duck failure gets resolved. even i tried to move the impl DeviceChannelRequest from device_operator.rs to thunder_async_client.rs file but getting error like error[E0116]: cannot define inherent `impl` for a type outside of the crate where the type is defined.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: changed to response.id instead of None

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
nnaveen979: done.

core/main/src/broker/endpoint_broker.rs
nnaveen979: reverted the change

device/thunder_ripple_sdk/src/bootstrap/setup_thunder_pool_step.rs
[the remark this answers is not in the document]
nnaveen979: done

core/sdk/src/api/device/device_operator.rs
nnaveen979: this is the black duck report link : https://gist.github.com/rdkcmf-jenkins/8ecec0494b4aa4825667f39f2ac2a2d4

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,563 @@
...
+        tokio::pin! {
+            let read = new_wrx.next();
+        }
+
+        tokio::select! {
+            Some(value) = &mut read => {
+                match value {
+                    Ok(v) => {
+                        if let tokio_tungstenite::tungstenite::Message::Text(t) = v {
+                            if self.status_manager.is_controller_response(self.get_sender(), callback.clone(), t.as_bytes()).await {
```
pahearn73, line 254: Added this since status manager requests are also being sent over this channel.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: done.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: done

core/sdk/src/api/device/device_operator.rs
nnaveen979: i tried to move the complete device_operator.rs file to into thunder_ripple_sdk-->client--> but in the last i found a lot of clippy error's.

nnaveen979: which are not related my changes but still getting the clippy failures.

nnaveen979: these clippy errors occurs in different submodules.

nnaveen979: @pahearn73  i feel its better to not move this file from existing directory.

nnaveen979: device_operator.rs file moved to thunder_ripple_sdk directory.

nnaveen979: file moved to thunder_ripple_sdk directory

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: @Vinodsathyaseelan , i have gone through the refactored functions in thunder_broker.rs check_and_generate_plugin_activation_request() and prepare_request() is there any specific reason for refactoring.  please let me. know.
in thunder_async_client.rs --> prepare_request() has the same logic.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,560 @@
...
+
+    pub fn new(callback: AsyncCallback, sender: AsyncSender) -> Self {
+        Self {
+            status_manager: StatusManager::new(),
+            sender,
+            callback,
+        }
+    }
+
+    pub async fn process_new_req(&self, request: String, url: String, callback: AsyncCallback) {
```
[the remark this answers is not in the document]
nnaveen979: websocket close change is added in fn process_new_req()

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,566 @@
...
+                },
+                Some(request) = tr.recv() => {
+                    debug!("Got request from receiver for thunder {:?}", request);
+                    // here prepare_request will check the plugin status and add json rpc format
+                    match client_c.prepare_request(&request) {
+                        Ok(updated_request) => {
+                            debug!("Sending request to thunder {:?}", updated_request);
+                            for r in updated_request {
+                                //check if it's subscription call use the already existing websocket for sending
+                                if r.contains(".register") {
```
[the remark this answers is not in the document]
nnaveen979: 1.point 1 is clear changes done , tested and committed, 
2.point 2 : after sending the ws_tx request, we will be in ws_rx for response form thunder here my doubt is do we need to skip the controller response checks and invoke the Self::handle_jsonrpc_response()?  but in my observation i feel like these controller response checks happens in the ripple boot up. im not clear on this point.

device/thunder_ripple_sdk/src/client/device_operator.rs
```diff
@@ -37,13 +38,46 @@ pub trait DeviceOperator: Clone {
...
+                let method = collection.pop().unwrap_or_default();
+                let callsign = collection.join(".");
+                (callsign, method.into())
+            }
+            DeviceChannelRequest::Subscribe(s) => (s.module.clone(), s.event_name.clone()),
+            DeviceChannelRequest::Unsubscribe(u) => (u.module.clone(), u.event_name.clone()),
+        }
+    }
+
+    pub fn is_subscription(&self) -> bool {
```
pahearn73, line 68: I don't think `is_subscription()` and `is_unsubscribe()` are used anywhere, if not can we remove?

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,565 @@
...
+        }
+    }
+
+    pub async fn start(
+        &self,
+        url: &str,
+        mut tr: Receiver<ThunderAsyncRequest>,
+    ) -> Receiver<ThunderAsyncRequest> {
+        let callback = self.callback.clone();
+        let (mut ws_tx, mut ws_rx) = Self::create_ws(url).await;
```
pahearn73, line 283: With the original websocket pool, if the websocket(s) become disconnected they reconnect and re-register any subscriptions, it doesn't look like we're doing either of those things?

device/thunder_ripple_sdk/src/client/thunder_client.rs
```diff
@@ -15,53 +15,103 @@
...
+    ) {
+        let client_c = client.clone();
+
+        tokio::spawn(async move {
+            if let Some(thndr_asynclient) = &client_c.thunder_async_client {
+                thndr_asynclient
+                    .start(&thndr_endpoint_url, request_tr)
+                    .await;
+            }
+            error!("Thunder disconnected so reconnecting");
```
pahearn73, line 78: Remove this log, no attempts to reconnect occur after this. Maybe ThunderAsyncClient::start should be responsible for socket reconnects.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: @pahearn73 correct me if i wrong , 
1. websocket disconnected means are we supposed to check the connection status in the loop of thunder_async_client::start fn(). 
2. if the connection disconnected means we need to re-establish the web socket again with in the same thread?
3. after re-establishing the websocket we have to send the same subscribe request made during the connection failure? 
4. i have a question that suppose i made 4 subscription requests and then connection failed then when i reconnect the websocket do i have to make those 4 subscribe requests again?

device/thunder_ripple_sdk/src/client/thunder_client.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
pahearn73: Update: Per phone conversation will add resubscription map and logic.

pahearn73 (dismissed):

core/sdk/src/utils/rpc_utils.rs
```diff
@@ -20,3 +20,13 @@ use jsonrpsee::core::Error;
 pub fn rpc_err(msg: impl Into<String>) -> Error {
     Error::Custom(msg.into())
 }
+
+pub fn extract_tcp_port(url: &str) -> Result<String, Error> {
+    let url_split: Vec<&str> = url.split("://").collect();
+    if let Some(domain) = url_split.get(1) {
+        let domain_split: Vec<&str> = domain.split('/').collect();
+        Ok(domain_split.first().unwrap().to_string())
```
Vinodsathyaseelan, line 28: avoid unsafe `unwrap()`

Vinodsathyaseelan: As pointed out in CodeRabbit feedback, this function will fail for URL like `example.com:8080`

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -21,25 +21,120 @@ use crate::{
...
     plugin_param: ThunderPluginBootParam,
 ) -> Option<ThunderBootstrapStateWithClient> {
-    info!("Booting thunder");
-    if let Ok(state) = ThunderGetConfigStep::setup(state, plugin_param).await {
+    info!("Booting thunder initiated");
+    //by default enabling the thunderBroker
+    let state = if ext_client.get_bool_config("use_with_thunder_async_client") {
+        info!("Using thunder_async_clinet");
+        let mut extn_client = ext_client.clone();
+        let mut gateway_url = url::Url::parse(GATEWAY_DEFAULT).unwrap();
```
Vinodsathyaseelan, line 58: Use `unwrap_or_else(|_|` here

device/thunder_ripple_sdk/src/client/device_operator.rs
```diff
@@ -116,6 +138,30 @@ impl DeviceResponseMessage {
...
+        sub_id: Option<String>,
+    ) -> Option<DeviceResponseMessage> {
+        let mut device_response_msg = None;
+        if let Some(res) = &json_resp.result {
+            device_response_msg = Some(DeviceResponseMessage::new(res.clone(), sub_id));
+        } else if let Some(er) = &json_resp.error {
+            device_response_msg = Some(DeviceResponseMessage::new(er.clone(), sub_id));
+        } else if json_resp.clone().method.is_some() {
+            if let Some(params) = &json_resp.params {
+                let dev_resp = serde_json::to_value(params).unwrap();
```
Vinodsathyaseelan, line 157: Avoid unsafe `unwrap()` here.
Try to use 
`if let Ok(dev_resp) = serde_json::to_value(params) `

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,589 @@
...
+                    "create_ws: endpoint {} failed with retry for last {} secs in {}",
+                    endpoint, index, tcp_port
+                );
+            }
+            index += 1;
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+    }
+
+    fn prepare_request(&self, request: &ThunderAsyncRequest) -> Result<String, RippleError> {
```
Vinodsathyaseelan, line 144: We have already moved the plugin status checking and activation logic out of `prepare_request` function. 
Please refer https://github.com/rdkcentral/Ripple/blob/main/core/main/src/broker/thunder_broker.rs#L472

device/thunder_ripple_sdk/src/tests/contracts/thunder_controller_pacts.rs
```diff
@@ -1,6 +1,10 @@
 use crate::{
     client::{
-        plugin_manager::ThunderActivatePluginParams, thunder_client::ThunderClient,
+        device_operator::{
```
Vinodsathyaseelan, line 3: What is the status of RDK contract testing after these changes ? Will the tests still be able to generate RDK contract with these modifications?

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -0,0 +1,589 @@
...
+        Self {
+            status_manager: StatusManager::new(),
+            sender,
+            callback,
+            subscriptions: HashMap::new(),
+        }
+    }
+
+    pub async fn process_new_req(&mut self, request: String, url: String) {
+        let (mut ws_tx, mut ws_rx) = Self::create_ws(&url).await;
```
Vinodsathyaseelan, line 234: Do we really need to `create_ws` on every request ?

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
nnaveen979: here GATEWAY_DEFAULT is a const, so getting error in this scenario is not applicable.
but i tried to avoid the unwrap() but i'm supposed write like i mentioned below 

        let mut gateway_url = match url::Url::parse(GATEWAY_DEFAULT) {
            Ok(url) => url,
            Err(err) => {
                warn!(
                    "Could not parse default gateway URL '{}': {}",
                    GATEWAY_DEFAULT, err
                );
                url::Url::parse("ws://127.0.0.1:8088/jsonrpc").unwrap()
            }
        };

        let mut gateway_url = url::Url::parse(GATEWAY_DEFAULT).unwrap_or_else(|err| {
            warn!("Could not parse default gateway URL '{}': {}", GATEWAY_DEFAULT, err);
            url::Url::parse("ws://127.0.0.1:8088/jsonrpc").unwrap()
        });

in the above two approaches i have to use unwrap again. i keep the line as it is like:
let mut gateway_url = url::Url::parse(GATEWAY_DEFAULT).unwrap();

device/thunder_ripple_sdk/src/client/device_operator.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: No,  removed the new websocket for new request. make use of same web socket which created earlier.

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
nnaveen979: two functions separated now these functions are more readable

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
nnaveen979: done

device/thunder_ripple_sdk/src/tests/contracts/thunder_controller_pacts.rs
nnaveen979: all tests are passing

core/sdk/src/utils/rpc_utils.rs
nnaveen979: done

device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -194,37 +194,46 @@ impl ThunderAsyncClient {
                 .generate_plugin_activation_request(callsign.clone());
             return Ok(request.to_string());
         }
+
         // Generate the appropriate JSON-RPC request based on the type of DeviceChannelRequest
```
pahearn73, line 198: RELATED: We seem to do a lot of en/decoding of request data across different types that get queued, we should take another look and see if that can get cut-down a little, doesn't seem like we're storing things in the most ideal format? Not now but something for later...

B-Larsen (approved):

Vinodsathyaseelan (approved):

PR #677, opened 2024-11-11
feat: Support for Firebolt rpc 2.0

satlead:
## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/provider_broker_state.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+    pub fn check_provider_request(
+        &self,
+        request: &RpcRequest,
+        permission: &Vec<FireboltPermission>,
+        session: Option<Session>,
+    ) -> Option<ProviderResult> {
+        debug!("Method {}", request.method);
+        if request.method.contains(".provide") {
```
pahearn73, line 62: Is there a better way to determine this instead of mandating that a method starts with `.provide`, e.g. can we look up a tag in the schema?

satlead: Its a combination of .provide and a firebolt capability with a role of Provides
`if let Some(p) = Self::get_permission(permission) {`

pahearn73: Maybe in another iteration we can use the RPC validator to query for the tag.

pahearn73 (approved):

PR #679, opened 2024-11-12
OpenRPC Validator: Not always returning method result schema

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/openrpc_validator/src/lib.rs
+++ b/openrpc_validator/src/lib.rs
@@ -130,6 +130,9 @@ impl FireboltOpenRpc {
         if let Some(method) = self.get_method_by_name(name) {
             if let Some(result_schema_map) = method.result.schema.as_object() {
                 if let Some(any_of_map) = result_schema_map.get("anyOf") {
+                    // Iterate the anyOf type array and get the first one that matches. With the current firebolt APIs it will happen
+                    // to be the correct type, but this is extremely fragile and should be addressed in future firbolt spec revisions.
+                    // Ripple needs a way to determine the explicit result type.
                     if let Some(any_of_array) = any_of_map.as_array() {
                         for value in any_of_array.iter() {
                             if let Some(result_properties_map) =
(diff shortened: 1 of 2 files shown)


openrpc_validator/src/lib.rs
```diff
@@ -115,23 +115,39 @@ impl FireboltOpenRpc {
...
+        let result_value = result_schema_map.get("type").unwrap_or(&Value::Null);
+        result_map.insert(rpc_method.result.name.clone(), result_value.clone());
+        result_map
+    }
+
     pub fn get_result_properties_schema_by_name(&self, name: &str) -> Option<Map<String, Value>> {
         if let Some(method) = self.get_method_by_name(name) {
             if let Some(result_schema_map) = method.result.schema.as_object() {
                 if let Some(any_of_map) = result_schema_map.get("anyOf") {
+                    // FIXME: This will always return the result schema for the first object in the anyOf array, because we don't have a way to determine
```
pahearn73, line 139: With the current firebolt schemas we never get into this situation where we need to access nested result schemas like this, but still needs more discussion on how we can support this in the future. Had prior discussions related to this with KP and the conclusion is that it's something that needs figuring out. I'll re-open that dialog.

openrpc_validator/src/lib.rs
```diff
@@ -115,23 +115,39 @@ impl FireboltOpenRpc {
...
+                        return Some(self.resolve_schema_properties(
+                            any_of_array[0].as_object().unwrap(),
+                            &method,
+                        ));
                     } else {
                         return None;
                     }
                 } else {
-                    return self.get_result_ref_schemas(result_schema_map);
+                    return Some(self.resolve_schema_properties(result_schema_map, &method));
```
pahearn73, line 150: This is the actual line that was causing the problem. Previously it assumed the result always contained a $ref to an object but there are cases where the result isn't an object but just a simple type like a string. This new method `resolve_schema_properties()` accounts for that.

brendanobra (approved):

B-Larsen (approved):

PR #687, opened 2024-11-21
Ripple provider app-passthrough: Result wrapped in result name field from schema

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/provider_registrar.rs
+++ b/core/main/src/firebolt/handlers/provider_registrar.rs
@@ -44,7 +44,7 @@ use ripple_sdk::{
         },
         gateway::rpc_gateway_api::{CallContext, CallerSession},
     },
-    log::{error, info, warn},
+    log::{error, info},
     tokio::{sync::oneshot, time::timeout},
     utils::serde_utils::SerdeClearString,
 };
@@ -296,7 +296,7 @@ impl ProviderRegistrar {
             let mut params_sequence = params.sequence();
             let call_context: Option<CallContext> = params_sequence.next().ok();
 
-            let mut event_data: Value = match params_sequence.next() {
+            let event_data: Value = match params_sequence.next() {
                 Ok(r) => r,
                 Err(e) => {
                     error!("callback_app_event_emitter: Error: {:?}", e);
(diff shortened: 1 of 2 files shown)


core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -296,73 +296,69 @@ impl ProviderRegistrar {
...
             let call_context: Option<CallContext> = params_sequence.next().ok();
 
-            let mut event_data: Value = match params_sequence.next() {
+            let event_data: Value = match params_sequence.next() {
                 Ok(r) => r,
                 Err(e) => {
                     error!("callback_app_event_emitter: Error: {:?}", e);
                     return Err(Error::Custom("Missing event_data".to_string()));
                 }
             };
```
pahearn73, line 306: Unrelated but cleaned-up this block a bit to also support events that contain a "non-object" payload, if we ever need it.

core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -296,73 +296,69 @@ impl ProviderRegistrar {
...
-                            &Value::Object(result_map),
-                        )
-                        .await;
+                        event_data.clone()
                     }
-                } else {
-                    error!("callback_app_event_emitter: Result schema not found");
-                    return Err(Error::Custom(String::from("Result schema not found")));
                 }
+                _ => event_data.clone(),
```
pahearn73, line 343: If the incoming event data is not a Value::Object then just emit whatever it is, no need to check for app ID injection.

core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -499,6 +495,9 @@ impl ProviderRegistrar {
                                             }
                                         }
                                         return Ok(Value::Object(response_map));
+                                    } else {
+                                        // Method returns a non-object type, just return it.
+                                        return Ok(provider_response_value);
```
pahearn73, line 500: Previously the OpenRPC Validator would return the name of the result field as an actual field in the result, causing the problem. That's now corrected in the validator, and here if we don't get a list of properties from the schema (because it's a non-object) in order to do app ID injection, we just return the actual response value (in this case a `Value::String`)

openrpc_validator/src/lib.rs
```diff
@@ -115,17 +117,6 @@ impl FireboltOpenRpc {
         None
     }
 
-    fn get_result_property(
```
pahearn73, line 118: This is what was generating a property map out of the result's name field in the schema, we don't want to do that. If there are no result properties just return `None`.

brendanobra (approved):

Nv-Delle1416 (approved):

pahearn73:
Closing due to difficult conflicts, creating new PRs.

PR #690, opened 2024-11-25
Ripple provider app-passthrough: Result wrapped in result name field from schema

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/handlers/provider_registrar.rs
+++ b/core/main/src/firebolt/handlers/provider_registrar.rs
@@ -44,7 +44,7 @@ use ripple_sdk::{
         },
         gateway::rpc_gateway_api::{CallContext, CallerSession},
     },
-    log::{error, info, warn},
+    log::{error, info},
     tokio::{sync::oneshot, time::timeout},
 };
 use serde_json::{Map, Value};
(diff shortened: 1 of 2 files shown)


core/main/src/firebolt/handlers/provider_registrar.rs
```diff
@@ -292,61 +292,59 @@ impl ProviderRegistrar {
...
+                                    result_map.insert(key.clone(), Value::Null);
+                                }
                             } else {
-                                error!("callback_app_event_emitter: Missing call context, could not determine app ID");
-                                result_map.insert(key.clone(), Value::Null);
-                            }
-                        } else {
-                            error!(
+                                error!(
                                 "callback_app_event_emitter: Missing field in event data: field={}",
```
brendanobra, line 327: Should this also return `Err<T>` ?

pahearn73: No, because we still want to return the event_handler and emit the event, but just log that the event data isn't what was expected.

brendanobra (approved):

Nv-Delle1416 (approved):

PR #693, opened 2024-11-27
fix: Multiple instances of Thunder Broker created during Ripple bootstrap

Vinodsathyaseelan:
## What

 Skipped Thunder Endpoint creation when setting up OtherBrokers.

## Why

ThunderBroker instances are created at multiple places. 

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -458,6 +458,10 @@ impl EndpointBrokerState {
 
     pub fn build_other_endpoints(&mut self, session: Option<AccountSession>) {
         for (key, endpoint) in self.rule_engine.rules.endpoints.clone() {
+            // skip thunder endpoint as it is already built using build_thunder_endpoint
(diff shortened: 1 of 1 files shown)


brendanobra (approved):

core/main/src/broker/endpoint_broker.rs
```diff
@@ -458,6 +458,10 @@ impl EndpointBrokerState {
 
     pub fn build_other_endpoints(&mut self, session: Option<AccountSession>) {
         for (key, endpoint) in self.rule_engine.rules.endpoints.clone() {
+            // skip thunder endpoint as it is already built using build_thunder_endpoint
+            if key == "thunder" {
```
pahearn73, line 462: Maybe this should instead be:

`if let RuleEndpointProtocol::Thunder = endpoint.protocol {`

Vinodsathyaseelan: Agreed.  Modified the code as suggested.

PR #702, opened 2024-12-10
mock_device: numerous fixes

brendanobra:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


device/mock_device/src/mock_web_socket_server.rs
```diff
@@ -106,20 +107,78 @@ type WSConnection = Arc<Mutex<HashMap<String, SplitSink<WebSocketStream<TcpStrea
...
+struct ApiStat {
+    method: String,
+    count: u32,
+}
+#[derive(Debug, Serialize, Deserialize)]
+struct ApiStats {
+    stats: Vec<ApiStat>,
+    total: u32,
+}
+impl StatsCollector {
```
brendanobra, line 137: This is little easter egg: since we have a control point at the thunder level in mock_device, seemed like it was useful/interesting to collect some stats - this will create a histogram if all thunder calls that are serviced by mock_device. The histogram is put in `stats.json` in the eos-ripple directory when (and only when) mock_device is used

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -655,18 +656,29 @@ impl ExtnClient {
         if !self.sender.get_cap().is_main() {
             return Err(RippleError::InvalidAccess);
         }
+        debug!("Main internal request with payload: {:?}", payload);
```
Vinodsathyaseelan, line 659: Do we need these DEBUG level logs ? Will this contribute to log bloating ?

brendanobra: Updated to trace, although these would only be for non prod builds

maggie98choy (approved):

pahearn73 (approved):

B-Larsen (approved):

PR #710, opened 2025-01-12
Improved Thunder logging, general logging improvements

brendanobra:
## What

Ostensibly, this PR addresses thunder request/response logging being remove awhile ago.  It also goes quite a bit further to add contextual (app_id, session id, etc.) log metadata enrichment.

All changes are using generics, so are compile time only, and are reasonably runtime efficient, but do introduce a 16KB binary image overhead (this was divined by compiling release on the feature branch  vs same on main , both on x86_64 linux)

Also note that I am running the code in this PR on my fork against coderabbit.ai: https://github.com/brendanobra/Ripple/pull/14

## Why
Support/triage
## How
This PR introduces:
1) LogSignal.rs , which attempts to provide a convenient, low overhead API surface for support/triage logging (i.e. NOT developer logging)
2) From implementation for CallContext, BrokeRequest, etc. to provide rust idiomatic and leveraged metadata enrichment of the log outputs in question
## Test
See unit tests for Logsignal. Also tested manually.
## Checklist

- [x ] I have self-reviewed this PR
- [ x] I have added tests that prove the feature works or the fix is effective


bsenth200 (approved):

core/main/src/broker/http_broker.rs
```diff
@@ -123,9 +123,12 @@ impl EndpointBroker for HttpBroker {
         let (tx, mut tr) = mpsc::channel(10);
         let broker = BrokerSender { sender: tx };
         let client = Client::new();
+
         let _ =  endpoint.get_url().parse().map_err(|e| error!("broker url {:?} in endpoint is invalid, cannot start http broker. error={}",endpoint,e) ).map(|uri| tokio::spawn(async move {
             while let Some(request) = tr.recv().await {
                 debug!("http broker received request={:?}", request);
+                LogSignal::new("http_broker".to_string(), "starting".to_string(), request.rpc.ctx.clone())
```
pahearn73, line 130: I see this logs "starting", should this be outside of the loop?

pahearn73 (approved):

core/main/src/broker/http_broker.rs
brendanobra: The intent is to log that it is starting to process a new request -  i have cleaned up the message to be less vague :)

Nv-Delle1416 (approved):

PR #714, opened 2025-01-15
feat: composite requet to support end point error response

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -1026,15 +1026,18 @@ impl BrokerOutputForwarder {
                             // Apply response rule using params if there is any; otherwise, apply response rule using main broker request's response rule
                             let mut apply_response_using_main_req_needed = true;
                             if let Some(params) = output.data.params {
-                                for (key, value) in params.as_object().unwrap() {
-                                    if key == "response" {
-                                        let filter = value.as_str().unwrap().to_string();
-                                        apply_response_using_main_req_needed = false;
-                                        apply_response(
-                                            filter,
-                                            &rpc_request.ctx.method,
-                                            &mut response,
-                                        );
+                                if let Some(param) = params.as_object() {
+                                    for (key, value) in param {
+                                        if key == "response" {
+                                            if let Some(filter) = value.as_str() {
+                                                apply_response_using_main_req_needed = false;
+                                                apply_response(
+                                                    filter.to_string(),
+                                                    &rpc_request.ctx.method,
+                                                    &mut response,
+                                                );
+                                            }
+                                        }
                                     }
                                 }
                             }
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -828,7 +828,7 @@ pub trait EndpointBroker {
     /// client for consumption
     fn handle_jsonrpc_response(
         result: &[u8],
-        callback: BrokerCallback,
+        callback: BrokerCallback, _params: Option<Value>,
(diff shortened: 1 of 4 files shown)


core/main/src/broker/endpoint_broker.rs
```diff
@@ -878,12 +878,31 @@ impl BrokerOutputForwarder {
...
-                            ) {
-                                apply_response(filter, &rpc_request.ctx.method, &mut response);
-                            } else if response.result.is_none() && response.error.is_none() {
-                                response.result = Some(Value::Null);
+                            // Apply response rule using params if there is any; otherwise, apply response rule using main broker request's response rule
+                            let mut apply_response_using_main_req_needed = true;
+                            if let Some(params) = output.data.params {
+                                for (key, value) in params.as_object().unwrap() {
+                                    if key == "response" {
+                                        let filter = value.as_str().unwrap().to_string();
```
brendanobra, line 886: avoid unwrap if possible - this is very unlikely to fail, but if it does, would be bad

core/main/src/broker/thunder_broker.rs
```diff
@@ -100,6 +123,28 @@ impl ThunderBroker {
...
 
+    // This function is used to start a timer and when the timer is up then purge composite request
+    fn start_purge_composite_request_timer(
+        composite_request_list: Arc<Mutex<HashMap<u64, RpcRequest>>>,
+        id: u64,
+    ) {
+        debug!("Start composite request purge timer");
+        let now = SystemTime::now();
+        // start a new thread to purge the composite request after 8 seconds
+        tokio::spawn(async move {
```
Vinodsathyaseelan, line 134: Will this thread be spawned for each outstanding request or only one timer in case any request outstanding. There should be only one thread which manages the timeout of all outstanding requests. 

What happens when the request get timed out ? Will the individual response rule apply as not success ?

core/main/src/broker/thunder_broker.rs
```diff
@@ -146,7 +192,30 @@ impl ThunderBroker {
                                     else {
                                         // send the incoming text without context back to the sender
                                         let id = Self::get_id_from_result(t.as_bytes());
-                                        Self::handle_jsonrpc_response(t.as_bytes(),broker_c.get_broker_callback(id).await)
+
+                                        /* Retrieve composite params and add it to JsonRpcApiResponse */
+                                        let rpc_req = broker_c.get_composite_request(id).await;
```
Vinodsathyaseelan, line 197: It might be more readable if you move this logic to a separate function.

core/main/src/broker/thunder_broker.rs
maggie98choy: Fixed only one thread manage purging outstanding request.

Unfortunately If a request get timed out, individual response rule won't apply.

core/main/src/broker/thunder_broker.rs
```diff
@@ -100,6 +144,49 @@ impl ThunderBroker {
         self.default_callback.clone()
     }
 
+    // Start a timer to purge individual composite request that are older than 8 seconds
+    fn start_purge_composite_request_timer(&self) {
+        let composite_request_list = self.composite_request_list.clone();
+        let mut interval = time::interval(Duration::from_millis(3000));
+        let purge_thread_started = self.composite_request_purge_started.clone();
+        tokio::spawn(async move {
+            if *purge_thread_started.lock().await {
```
Vinodsathyaseelan, line 153: Can we do this check even before spawning the thread?

Vinodsathyaseelan (approved):

tfarz123 (approved):

PR #715, opened 2025-01-15
Create ExtnBroker and RPCRequestProcessor, add RPCRouter to Ripple SDK & impl wildcard rules support

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/endpoint_broker.rs
```diff
@@ -393,7 +394,7 @@ impl EndpointBrokerState {
         }
     }
 
-    fn get_extn_message(&self, id: u64, is_event: bool) -> Result<ExtnMessage, RippleError> {
+    pub fn get_extn_message(&self, id: u64, is_event: bool) -> Result<ExtnMessage, RippleError> {
```
satlead, line 397: Does it need to be Public

core/main/src/broker/extn_broker.rs
```diff
@@ -0,0 +1,143 @@
...
+    ) -> BrokerSender {
+        let (tx, mut rx) = mpsc::channel::<BrokerRequest>(10);
+
+        tokio::spawn(async move {
+            while let Some(broker_request) = rx.recv().await {
+                trace!("Received message {:?}", broker_request);
+                let rpc_request = broker_request.rpc.clone();
+                let rule = broker_request.rule.clone();
+                let alias = rule.alias;
+                let id = if alias == "ripple:channel:gateway:badger" {
```
satlead, line 53:
```
"some.*": {
      "endpoint": "extn",
      "alias": "ripple:channel:gateway:some"
    },
```
Alias should reference to the Symbol identifier

core/sdk/src/extn/extn_id.rs
```diff
@@ -295,6 +295,20 @@ impl ExtnId {
...
 
+    /// Checks if the given capability is a badger channel.
+    /// # Examples
+    /// ```
+    /// use ripple_sdk::extn::extn_id::{ExtnId,ExtnClassId};
+    ///
+    /// let badger_channel = ExtnId::new_channel(ExtnClassId::Gateway, "badger".into());
+    /// assert!(badger_channel.is_badger_channel());
+    /// ```
+    pub fn is_badger_channel(&self) -> bool {
```
pahearn73, line 306: Not sure where this is needed but can we do checks without using bespoke code like this? E.g. by parsing the ID etc.?

core/main/src/broker/endpoint_broker.rs
bsenth200: Reverted change.

pahearn73 (approved):

core/main/src/broker/endpoint_broker.rs
```diff
@@ -502,18 +504,18 @@ impl EndpointBrokerState {
             BrokerRequest::new(&rpc_request_c, rule, workflow_callback),
         )
     }
-    pub fn build_thunder_endpoint(&mut self) {
+    pub fn build_thunder_endpoint(&mut self, _ps: PlatformState) {
```
satlead, line 507: why do we need a new parameter in this method when it is not used?

core/main/src/broker/endpoint_broker.rs
```diff
@@ -659,11 +667,13 @@ impl EndpointBrokerState {
...
             .emit_debug();
         }
+
         trace!("found rule {:?}", found_rule);
+
         if found_rule.is_some() {
             let rule = found_rule.unwrap();
 
-            if rule.alias == "static" {
+            if rule.alias == *"static" {
```
satlead, line 676: is the * intended?

PR #721, opened 2025-02-06
set migration status to true for use case of closed caption settings …

maggie98choy:
…has not been previously used in legacy storage

## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/thunder/user_data_migrator.rs
```diff
@@ -285,12 +279,14 @@ impl UserDataMigrator {
         let request_id = EndpointBrokerState::get_next_id();
         let call_sign = "org.rdk.PersistentStore.1.".to_owned();
 
+        let (response_tx, response_rx) = mpsc::channel(16);
```
Vinodsathyaseelan, line 282: You may use a oneshot::channel here.

Nv-Delle1416 (approved):

PR #725, opened 2025-02-10
Added Ripple call context for all Ripple logs to allow from proper tracing of call flows 

bsenth200:
…per tracing of call flows

## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/endpoint_broker.rs
```diff
@@ -821,12 +821,33 @@ pub trait EndpointBroker {
                             format!("{}_request", rpc_request.rpc.ctx.method),
                         );
                     }
+                    LogSignal::new(
```
pahearn73, line 824: Do we also need a new LogSignal in the `if let Some(filter) = rpc_request
                        .rule
                        .transform
                        .get_transform_data(super::rules_engine::RuleTransformType::Request)` block, above?

brendanobra (approved):

Nv-Delle1416 (approved):

PR #727, opened 2025-02-13
Added firebolt telemetry support for rule-based firebolt requests

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/firebolt/firebolt_gateway.rs
```diff
@@ -146,6 +149,61 @@ impl FireboltGateway {
...
+            trace!("handle_broker_callback: resp={:?}", resp);
+
+            match resp {
+                Ok(has_broker_output) => {
+                    if let Some(broker_output) = has_broker_output {
+                        let now = Utc::now().timestamp_millis();
+
+                        let mut api_message = ApiMessage::new(
+                            protocol,
+                            serde_json::to_string(&broker_output.data).unwrap_or_default(),
```
Vinodsathyaseelan, line 180: Is it better to handle the errors here and log those instead of going ahead with some default values ?

core/main/src/firebolt/firebolt_gateway.rs
```diff
@@ -278,11 +336,18 @@ impl FireboltGateway {
                         }
                     }
 
-                    if !platform_state.endpoint_state.handle_brokerage(
+                    let requestor_callback_tx =
+                        Self::handle_broker_callback(platform_state.clone(), request_c.clone());
```
Vinodsathyaseelan, line 340: Nit: You may pass these parameters by reference to avoid cloning.

Vinodsathyaseelan (dismissed):

core/main/src/firebolt/firebolt_gateway.rs
pahearn73: Need to clone as they are moved to another thread.

core/main/src/firebolt/firebolt_gateway.rs
pahearn73: Added, thanks.

brendanobra (approved):

B-Larsen (approved):

PR #731, opened 2025-02-18
RPPL-2649 throttling handling for the event handlers

nnaveen979:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


device/thunder_ripple_sdk/src/events/thunder_event_processor.rs
```diff
@@ -318,6 +326,27 @@ impl ThunderEventProcessor {
         }
         false
     }
+
+    pub fn add_backoff(&self, event_name: &str, previous_back_off: i32, back_off_value: i32) {
```
pahearn73, line 330: This is an object method (self), so can't you get rid of the `previous_back_off` parameter and do the setting of both `ThunderBackOff` fields in here? That way you don't have to do it externally and it's properly encapsulated.

pahearn73: Also, since it seems like `back_off_value` is explicit and relative to the current back-off, can we rename this method to `set_backoff()`?

device/thunder_ripple_sdk/src/thunder_state.rs
```diff
@@ -150,14 +151,17 @@ impl ThunderState {
             tokio::spawn(async move {
                 while let Some(request) = r.recv().await {
                     if let Some(id) = request.sub_id {
-                        let value = request.message.clone();
-                        if let Some(handler) = state_c.event_processor.get_handler(&id) {
+                        //check the back_off value
+                        if let Some(_back_off) = state_c.event_processor.get_backoff(&id) {
```
pahearn73, line 155: This could be `if state_c.event_processor.get_backoff(&id).is_some()`

device/thunder_ripple_sdk/src/thunder_state.rs
```diff
@@ -150,14 +151,17 @@ impl ThunderState {
...
+                            info!("back_off is not None. can't process event of ID:{}", id);
```
pahearn73, line 156: I wonder if we should make this a `warn!`.

device/thunder_ripple_sdk/src/thunder_state.rs
nnaveen979: done

device/thunder_ripple_sdk/src/events/thunder_event_processor.rs
nnaveen979: changes done

pahearn73 (approved):

Nv-Delle1416 (approved):

PR #733, opened 2025-02-19
feat: Update extn with caching

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/endpoint_broker.rs
```diff
@@ -555,7 +559,11 @@ impl EndpointBrokerState {
         result
     }
 
-    fn build_endpoint(&mut self, ps: Option<PlatformState>, request: BrokerConnectRequest) {
```
satlead, line 558: Do we need to update it to Ripple Client? What if other endpoint brokers need something else from Platform State

core/main/src/broker/endpoint_broker.rs
```diff
@@ -521,7 +521,11 @@ impl EndpointBrokerState {
         }
     }
 
-    pub fn build_other_endpoints(&mut self, ps: PlatformState, session: Option<AccountSession>) {
```
satlead, line 524: Same as https://github.com/rdkcentral/Ripple/pull/733/files#r1966006939

core/main/src/broker/endpoint_broker.rs
bsenth200: This is carried over from my previous badger changes pull request. I will revert this since we have now decided that PS is acceptable to pass.

core/main/src/broker/endpoint_broker.rs
bsenth200: This is carried over from my previous badger changes pull request. I will revert this since we have now decided that PS is acceptable to pass.

core/sdk/Cargo.toml
```diff
@@ -57,10 +58,10 @@ async-trait = "0.1.57"
 libloading = "0.7.4"
 async-channel = { version = "2.1.0", default-features = false }
 tree_magic_mini = { version = "3.0.3", optional = true}
+lazy_static = "1.5.0"
```
pahearn73, line 61: This is only used for test/mock, should this be under `dev-dependencies` instead?

bsenth200: Hi @pahearn73 , yeah, I tried moving lazy_static = “1.5.0” to dev-dependencies in sdk. However, when the unit tests are run from “ripple-badger-extn, ” they fail with the error: use of an undeclared crate or module `lazy_static.` So, I moved it back.

Even when the tests are run from “ripple-badger-extn,” I notice some discrepancies with SDK not being in test mode. This is the first time tests have been added in badger-extn. Additionally, one of the random tests fails inconsistently and passes when run again. We need to figure out why the SDK is not using the test cfg in this case. 

Let me know if I can fix it in any way. If not, I will create a ticket to resolve this issue and include an action in the badger-extn GitHub repository for unit test verification.

bsenth200: @pahearn73 Created new ticket - https://ccp.sys.comcast.net/browse/RPPL-3003 to add more unit tests in ripple-badger-extn, move lazy_static to dev-dependencies & create github action to enable unit test verification etc.

pahearn73 (dismissed):

brendanobra (approved):

satlead (approved):

B-Larsen (approved):

brendanobra (approved):

PR #744, opened 2025-03-02
feat: Create an RDK contract test framework using generic client.

Vinodsathyaseelan:
## What

Use a thin websocket client to send thunder call message to pact server instead of using thunder extension processor and Thunder client pool 
## Why

Many of the thunder extension processing functions are removed as Ripple is now using Rule Engine.
Rippple is migrating to AsyncThunderClient and will be removing the usage of Thunder ClientPool  
## How

## Test

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/tests/contracts/thunder_controller_pacts.rs
+++ b/device/thunder_ripple_sdk/src/tests/contracts/thunder_controller_pacts.rs
@@ -34,15 +34,19 @@ async fn test_register_state_change_event() {
         })
     );
 
-    let request = json!({
-        "jsonrpc": "2.0",
-        "id": 0,
-        "method": "Controller.1.register",
-        "params": json!({
-            "event": "statechange", "id": "client.Controller.1.events"
+    send_thunder_call_message!(
+        server_url.to_string(),
+        json!({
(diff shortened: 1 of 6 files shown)


device/thunder_ripple_sdk/src/tests/contracts/thunder_controller_pacts.rs
```diff
@@ -193,13 +157,12 @@ async fn test_display_settings_plugin_status() {
...
     );
 
-    let thunder_client = initialize_thunder_client(server_url.clone()).await;
-    let _resp: DeviceResponseMessage = perform_device_call_request(
-        thunder_client,
-        "Controller.1.status@org.rdk.DisplaySettings",
-        None,
-    )
-    .await;
+    let request = json!({
```
brendanobra, line 160: U could inline the request and make it even more compact concise: 
    send_thunder_call_message!(server_url.to_string(), json!({
        "jsonrpc": "2.0",
        "id": 3,
        "method": "Controller.1.status@org.rdk.DisplaySettings",
    })).await;

brendanobra (dismissed):

Nv-Delle1416 (approved):

brendanobra (approved):

PR #746, opened 2025-03-05
XIONE-16799: when thunderservice is not ready dont send response to caller

nnaveen979:
… user

## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/device/thunder_ripple_sdk/src/client/thunder_async_client.rs
+++ b/device/thunder_ripple_sdk/src/client/thunder_async_client.rs
@@ -185,7 +185,12 @@ impl ThunderAsyncClient {
         }
         // If the plugin is activating, return a service not ready error
         if status.state.is_activating() {
-            info!("Plugin {} is activating", callsign);
+            info!(
+                "Plugin {} is activating. Adding broker request to pending list",
+                callsign
+            );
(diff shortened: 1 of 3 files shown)


device/thunder_ripple_sdk/src/client/thunder_async_client.rs
```diff
@@ -367,13 +367,12 @@ impl ThunderAsyncClient {
                                 let response = ThunderAsyncResponse::new_error(request.id,e.clone());
                                 match e {
                                     RippleError::ServiceNotReady => {
-                                        info!("prepare request failed for request {:?}", request);
+                                        info!("Thunder Service not ready, request added to pending list{:?}", request);
```
pahearn73, line 370: Where is it being added? Or does it remain in the pending list upon error?

Vinodsathyaseelan: I believe there is still a bug in this area. The second request is not being added to the pending list.

nnaveen979:
closing this PR because actual PR details available in [RPPL-3030](https://ccp.sys.comcast.net/browse/RPPL-3030)

PR #751, opened 2025-03-13
Update handler code for Localization.addAdditionalInfo and Localization.removeAdditionalInfo to use internal JQ calls

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/endpoint_broker.rs
+++ b/core/main/src/broker/endpoint_broker.rs
@@ -901,53 +901,43 @@ pub trait EndpointBroker {
     /// Generic method which takes the given parameters from RPC request and adds rules using rule engine
     fn apply_request_rule(rpc_request: &BrokerRequest) -> Result<Value, RippleError> {
         if let Ok(mut params) = serde_json::from_str::<Vec<Value>>(&rpc_request.rpc.params_json) {
-            if params.len() > 1 {
-                if let Some(last) = params.pop() {
-                    if let Some(filter) = rpc_request
-                        .rule
-                        .transform
-                        .get_transform_data(super::rules_engine::RuleTransformType::Request)
-                    {
-                        let transformed_request_res = jq_compile(
-                            last,
-                            &filter,
-                            format!("{}_request", rpc_request.rpc.ctx.method),
-                        );
-
-                        LogSignal::new(
-                            "endpoint_broker".to_string(),
-                            "apply_request_rule".to_string(),
-                            rpc_request.rpc.ctx.clone(),
-                        )
-                        .with_diagnostic_context_item("success", "true")
-                        .with_diagnostic_context_item(
-                            "result",
-                            &format!("{:?}", transformed_request_res),
-                        )
-                        .emit_debug();
-
-                        return transformed_request_res;
-                    }
-                    LogSignal::new(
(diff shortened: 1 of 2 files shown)


core/main/src/broker/endpoint_broker.rs
```diff
@@ -901,53 +901,43 @@ pub trait EndpointBroker {
     /// Generic method which takes the given parameters from RPC request and adds rules using rule engine
     fn apply_request_rule(rpc_request: &BrokerRequest) -> Result<Value, RippleError> {
         if let Ok(mut params) = serde_json::from_str::<Vec<Value>>(&rpc_request.rpc.params_json) {
-            if params.len() > 1 {
```
pahearn73, line 904: There was a bug here where if there were no params to passed into the request, then we wouldn't process the Request transform at all. @satlead please confirm I understand your original intention WRT the request params.

pahearn73: These changes preserve the original logic of expecting `last` to be the second request parameter, if it exists at all.

core/main/src/firebolt/handlers/localization_rpc.rs
```diff
@@ -125,6 +129,61 @@ pub trait Localization {
...
+) -> RpcResult<()> {
+    match BrokerUtils::process_internal_main_request(
+        &mut platform_state,
+        "localization.additionalInfo",
+        None,
+    )
+    .await
+    {
+        Ok(Value::Object(additional_info_map)) => {
+            let mut info_map = additional_info_map.clone();
```
Vinodsathyaseelan, line 149: You may avoid cloning by using mutable `additional_info_map`.

`Ok(Value::Object(mut additional_info_map))`

core/main/src/firebolt/handlers/localization_rpc.rs
```diff
@@ -125,6 +129,61 @@ pub trait Localization {
...
+                        Value::String(set_map_entry_property.value.clone()),
+                    );
+                }
+                MapEntryProperty::Remove(remove_map_entry_property) => {
+                    info_map.remove(&remove_map_entry_property.key);
+                }
+            }
+
+            let params = Some(json!({
+                "value": serde_json::to_string(&info_map).unwrap(),
```
Vinodsathyaseelan, line 164: unsafe `unwrap()`

Vinodsathyaseelan (dismissed):

Vinodsathyaseelan (dismissed):

core/main/src/firebolt/handlers/localization_rpc.rs
```diff
@@ -48,7 +52,7 @@ use serde::Deserialize;
 #[derive(Deserialize, Debug)]
 pub struct SetMapEntryProperty {
     pub key: String,
-    pub value: String,
+    pub value: Value,
```
pahearn73, line 55: This fixes firebolt 1.5.0 support for string/int/bool values. Ticket includes request/responses to confirm fix and compatibility with any pre-existing stored additional info.

Vinodsathyaseelan (approved):

Rajanika (approved):

PR #753, opened 2025-03-13
Default logs signal to trace

pkumbh631:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/sdk/src/utils/logger.rs
```diff
@@ -44,6 +44,7 @@ pub fn init_logger(name: String) -> Result<(), fern::InitError> {
             ));
         })
         .level(filter)
+        .level_for("log-signal", log::LevelFilter::Trace)
```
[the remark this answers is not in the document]
pkumbh631: it's log_signal 'll update

Nv-Delle1416 (approved):

core/sdk/src/api/observability/log_signal.rs
```diff
@@ -187,7 +187,7 @@ impl<T: std::fmt::Display + ContextAsJson> LogSignal<T> {
         }
     }
     pub fn emit_debug(&self) {
-        log::debug!("{}", serde_json::Value::from(self));
```
pahearn73, line 190: @brendanobra : This was the change I recommended, but I thought you said you didn't want to do it this way. Change of heart or just a misunderstanding?

core/sdk/src/utils/logger.rs
pahearn73: So does this need to be changed or no?

core/sdk/src/utils/logger.rs
```diff
@@ -20,7 +20,7 @@ use std::{str::FromStr, sync::atomic::AtomicU32};
 pub static LOG_COUNTER: AtomicU32 = AtomicU32::new(1);
 
 pub fn init_logger(name: String) -> Result<(), fern::InitError> {
-    let log_string: String = std::env::var("RUST_LOG").unwrap_or_else(|_| "debug".into());
```
pahearn73, line 23: Do we want to default to `trace`? I thought the plan was to still default to `debug` but in situations where we wanted to display `log_signal` output we'd just set the log level to `trace`?

pkumbh631: 'll change it to debug, missed it

pkumbh631: @brendanobra There is only one ripple.service file for both prod and VBN, so do we want to default to INFO now?

brendanobra (approved):

core/sdk/src/utils/logger.rs
pahearn73: Why do we need to change it from what it was ("debug") at all? Wasn't VBN/PROD builds working fine as it was? And this isn't needed for the changes to log_signal is it?

core/main/src/bootstrap/manifest/device.rs
```diff
@@ -31,42 +29,3 @@ impl LoadDeviceManifestStep {
         r.expect("Need valid Device Manifest")
     }
 }
-
```
pahearn73, line 34: Did you mean to include these changes?

pahearn73: There's also more device manifest changes below...

core/sdk/src/api/manifest/device_manifest.rs
```diff
@@ -68,6 +68,8 @@ pub struct RippleConfiguration {
     pub metrics_logging_percentage: u32,
     #[serde(default)]
     pub internet_monitoring_configuration: InternetMonitoringConfiguration,
+    #[serde(default = "log_signal_default_level")]
+    pub log_signal_log_level: String,
```
pahearn73, line 72: I think if you make this a `log::LevelFilter` instead of a `String` then you don't have to do the to_str()/unwrap_or logic later.

core/main/src/bootstrap/manifest/device.rs
brendanobra: The load steps had to be moved from device_manifest to device to all the loading logic to be available to the initialization logic in logger. TL;DR:  intentional

PR #755, opened 2025-03-18
Feature/data governance cleanup

satlead:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/broker_utils.rs
```diff
@@ -72,6 +72,17 @@ impl BrokerUtils {
         }
     }
 
+    pub async fn process_for_app_main_request(
```
pahearn73, line 75: Maybe you don't have the latest from main but there is already a function, [process_internal_request()](https://github.com/rdkcentral/Ripple/blob/main/core/main/src/broker/broker_utils.rs#L83) that does this.

core/main/src/firebolt/firebolt_gateway.rs
```diff
@@ -291,11 +291,6 @@ impl FireboltGateway {
             .metrics
             .add_api_stats(&request_c.ctx.request_id, &request_c.method);
 
-        let metrics_timer = TelemetryBuilder::start_firebolt_metrics_timer(
```
pahearn73, line 294: I'm trying to understand the implications of re/moving the metrics timer logic. Doesn't this mean we're no longer accounting for all of the time spent processing requests, e.g. routing, permission checking, etc.?

core/main/src/firebolt/handlers/internal_rpc.rs
```diff
@@ -0,0 +1,115 @@
...
+
+use crate::{
+    firebolt::rpc::RippleRPCProvider,
+    service::{apps::app_events::AppEvents, telemetry_builder::TelemetryBuilder},
+    state::platform_state::PlatformState,
+    utils::rpc_utils::rpc_await_oneshot,
+};
+
+#[rpc(server)]
+pub trait Internal {
```
pahearn73, line 38: If we're defining a trait why don't we need it defined in the SDK instead of here?

core/main/src/processor/main_context_processor.rs
```diff
@@ -92,28 +79,7 @@ impl MainContextProcessor {
         {
             if let Some(session) = response.payload.extract() {
                 state.session_state.insert_account_session(session);
-                MetricsState::update_account_session(state).await;
                 event = CapEvent::OnAvailable;
-                let state_c = state.clone();
-                // update ripple context for token asynchronously
-                tokio::spawn(async move {
```
pahearn73, line 99: Guessing we're now doing this wherever `MetricsState` got moved to?

core/main/src/processor/main_context_processor.rs
```diff
@@ -145,110 +111,16 @@ impl MainContextProcessor {
...
-        debug!("handling internet connection change: {:?}", internet_state);
-        let internet_state = match internet_state {
-            Some(state) => state,
-            None => return,
-        };
-
-        if matches!(internet_state, InternetConnectionStatus::FullyConnected) {
-            // get internet monitoring interval from platformState configuration
-            let interval_in_seconds = state
-                .get_device_manifest()
```
pahearn73, line 242: A few questions:

1. If we're no longer maintaining the state of the internet connection, we're ok with waiting for the connection timeout however long that is? I thought we added state monitoring specifically to avoid that so wanted to make sure we're on the same page.
2. Did we also remove the monitoring interval and related fields from the device manifest?

core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -538,61 +534,80 @@ impl DelegatedLauncherHandler {
...
+            let params = serde_json::to_value(AppLifecycleStateChange {
+                app_id: app_id.to_owned(),
                 new_state: to_state.unwrap(),
-            });
+                previous_state: from_state,
+            })
+            .unwrap();
+            if BrokerUtils::process_for_app_main_request(
+                &mut self.platform_state.clone(),
+                "ripple.reportLifecycleStateChange",
```
pahearn73, line 579: What's the reasoning for sending things like this through the rules engine instead of just directly reporting in Ripple, in other words what are we doing differently across, say, Sky and Comcast to warrant needing rules?

core/sdk/src/api/default_storage_properties.rs
```diff
@@ -0,0 +1,235 @@
...
+    UnreconizedNamespace(String),
+    NotFound(String),
+}
+
+#[derive(Clone, Debug)]
+pub struct DefaultStorageProperties;
+
+impl DefaultStorageProperties {
+    pub fn get_bool(
+        state: &DefaultValues,
```
pahearn73, line 50: Guessing this is moved code but `state` should probably be `default_values` or similar since they do not change (state).

core/sdk/src/api/default_storage_properties.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+        } else {
+            Err(DefaultStoragePropertiesError::UnreconizedNamespace(
+                namespace.to_owned(),
+            ))
+        }
+    }
+
+    pub fn get_string(
+        state: &DefaultValues,
```
pahearn73, line 100: 'state1

core/sdk/src/api/default_storage_properties.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+        } else {
+            Err(DefaultStoragePropertiesError::UnreconizedNamespace(
+                namespace.to_owned(),
+            ))
+        }
+    }
+
+    pub fn get_number_as_u32(
+        state: &DefaultValues,
```
pahearn73, line 179: `state`

core/sdk/src/api/default_storage_properties.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+        } else {
+            Err(DefaultStoragePropertiesError::UnreconizedNamespace(
+                namespace.to_owned(),
+            ))
+        }
+    }
+
+    pub fn get_number_as_f32(
+        state: &DefaultValues,
```
pahearn73, line 212: `state`

core/sdk/src/framework/ripple_contract.rs
```diff
@@ -41,12 +41,7 @@ pub enum RippleContract {
     /// Used by Main application to provide internal contracts for Extensions
     #[default]
     Internal,
-    /// Provided by the distributor useful for adding Governance implementation for handling
-    /// privacy information and other sensitive data.
```
pahearn73, line 45: Curious why we're removing contract testing for these? Or are the contract enums gong to be defined elsewhere?

device/thunder_ripple_sdk/Cargo.toml
```diff
@@ -78,6 +79,7 @@ home = { version = "=0.5.5", optional = true }
 tree_magic_mini = { version = "=3.0.3", optional = true }
 rstest = { version = "0.18.2", optional = true, default-features = false }
 futures-util = { version = "0.3.28", features = ["sink", "std"], default-features = false, optional = true}
+chrono = { workspace = true, features = ["clock"]}
```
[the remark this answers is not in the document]
satlead: Merge issue cleared

core/sdk/src/framework/ripple_contract.rs
satlead: We are removing contracts which are no longer necessary for our implementation

core/main/src/service/apps/delegated_launcher_handler.rs
satlead: We are trying to remove contract driven Metrics to use Json RPC driven metrics. This would help us realign extensions to use Json Rpc instead of FFI

core/main/src/firebolt/handlers/internal_rpc.rs
satlead: It will be used internally within extensions not exposed as an open rpc to apps

core/main/src/firebolt/firebolt_gateway.rs
satlead: Observability was already removed which uses Timer, Counter this code was just setting these values

core/main/src/processor/main_context_processor.rs
satlead: Yes its moved to distributor extension as session manager

core/sdk/src/api/default_storage_properties.rs
satlead: Fixed

core/sdk/src/api/default_storage_properties.rs
satlead: Fixed

core/sdk/src/api/default_storage_properties.rs
satlead: Fixed

core/sdk/src/api/default_storage_properties.rs
satlead: Fixed

core/main/src/broker/broker_utils.rs
satlead: This method is a derivative method which has no call context so this method offers support for those instances where callcontext is not fully available
```
 async fn report_app_state_transition(
        &mut self,
        app_id: &str,
        method: &AppMethod,
        app_manager_response: Result<AppManagerResponse, AppError>,
    )
```

pahearn73 (dismissed):

B-Larsen (approved):

PR #760, opened 2025-03-20
feat: Added mock for unit tests

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


brendanobra (approved):

core/sdk/src/extn/client/extn_client.rs
```diff
@@ -741,6 +741,7 @@ impl ExtnClient {
         &mut self,
         payload: impl ExtnPayloadProvider,
         timeout_in_msecs: u64,
+        test_ctx: Option<String>,
```
pahearn73, line 744: It seems like there should be a way to mock (or add a test harness to) `ExtnClient` so that we don't have to  pass a test parameter into this method (I think it's called in 40+ places throughout Ripple), and maybe then be able to do away with the `#[cfg(all(not(feature = "mock"), not(test)))]` and `#[cfg(any(test, feature = "mock"))]` code blocks -- have you looked into doing something like that? So instead of having to call `get_next_mock_repsonse()` you would actually have some sort of mock `ExtnClient` that returns the mock responses.

B-Larsen (approved):

PR #762, opened 2025-03-21
Place Ripple Log Signal behind manifest flag

pkumbh631:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


bsenth200 (dismissed):

core/main/src/bootstrap/logging_bootstrap_step.rs
```diff
@@ -0,0 +1,25 @@
...
+    fn get_name(&self) -> String {
+        "LoggingBootstrapStep".into()
+    }
+
+    async fn setup(&self, state: BootstrapState) -> RippleResponse {
+        let device_manifest = &state.platform_state.device_manifest.configuration;
+
+        //set log signal log level
+        let log_signal_log_level = device_manifest.log_signal_log_level.clone();
+        set_log_level(&log_signal_log_level);
```
pahearn73, line 21: It looks like this will set the log level for all modules, not just `LogSignal`. In order to target just `LogSignal` I think you need to:

1. Create a `mod.rs` file in `Ripple/core/sdk/src/api/observability` similar to how it's done elsewhere.
2. Make the `LogSignal` module public by adding it to `Ripple/core/sdk/src/api/mod.rs`, see other `pub mod` entries there.
3. In your `setup()` code here, set the log level for the `LogSignal` mod similar to how it's done in `init_and_configure_logger()`, [here](https://github.com/rdkcentral/Ripple/blob/main/core/sdk/src/utils/logger.rs#L60).

pahearn73: I think that should work but needs testing.

brendanobra: Agree ^^^. The goal of this exercise is to decouple LogSignal initialization (as much as possible) from the logging system general initialization.

core/main/src/bootstrap/logging_bootstrap_step.rs
```diff
@@ -0,0 +1,43 @@
...
+#[async_trait]
+impl Bootstep<BootstrapState> for LoggingBootstrapStep {
+    fn get_name(&self) -> String {
+        "LoggingBootstrapStep".into()
+    }
+
+    async fn setup(&self, state: BootstrapState) -> RippleResponse {
+        //let device_manifest = &state.platform_state.device_manifest.configuration;
+        let manifest = state.platform_state.get_device_manifest();
+        let log_level = match manifest
```
pahearn73, line 20: I would rename this to `log_signal_level` just so it's clear it's not the log level in general.

pahearn73 (approved):

pahearn73 (approved):

core/sdk/src/utils/logger.rs
```diff
@@ -121,7 +114,17 @@ pub fn init_and_configure_logger(version: &str, name: String) -> Result<(), fern
         .level_for("tungstenite", log::LevelFilter::Off)
         .level_for("soketto", log::LevelFilter::Off)
         .level_for("tracing", log::LevelFilter::Off)
-        .chain(std::io::stdout())
-        .apply()?;
+        .chain(std::io::stdout());
```
pahearn73, line 117: Don't you have to call `chain()` after `if let Some(modules) = additional_modules {`?

core/main/src/bootstrap/logging_bootstrap_step.rs
```diff
@@ -0,0 +1,44 @@
...
+#[async_trait]
+impl Bootstep<BootstrapState> for LoggingBootstrapStep {
+    fn get_name(&self) -> String {
+        "LoggingBootstrapStep".into()
+    }
+
+    async fn setup(&self, state: BootstrapState) -> RippleResponse {
+        //let device_manifest = &state.platform_state.device_manifest.configuration;
+        let manifest = state.platform_state.get_device_manifest();
+        let log_signal_level = match manifest
```
pahearn73, line 20: Take a look at the Create Feature Flags of the log crate, there's a `serde` feature that allows the de/serialization of `LevelFilter`. If adding that feature doesn't add too much overhead, all of this can be removed and your entries in the manifest would just be `Info`, `Debug`, `Off`, etc.

core/sdk/src/api/observability/log_signal.rs
```diff
@@ -186,11 +187,34 @@ impl<T: std::fmt::Display + ContextAsJson> LogSignal<T> {
...
+                    log::trace!(target: target, "{}", message);
+                }
+                _ => {}
+            }
+        }
+    }
+
     pub fn emit_debug(&self) {
-        log::debug!("{}", serde_json::Value::from(self));
+        self.emit();
```
pahearn73, line 214: I don't think this is what we want to do, shouldn't this work similar to how the `info!()`, `debug!()`, etc. macros work, where, when called they simply don't log if the log level is insufficient? E.g. these would instead be:

```
pub fn emit_debug(&self) {
    log::debug!(target: target, "{}", message);
}

pub fn emit_info(&self) {
    log::info!(target: target, "{}", message);
}

pub fn emit_error(&self) {
    log::error!(target: target, "{}", message);
}

pub fn emit_trace(&self) {
    log::trace!(target: target, "{}", message);
}
```

pahearn73 (changes requested):

core/sdk/src/api/observability/log_signal.rs
brendanobra: My intent was to only need `emit(...)` and let the loglevel setting determine what level it outputs at - the caller should not really need to know about any of this  - this is different from regular developer logging

pahearn73: So we're never ever going to use log signal for tracing level stuff? Or errors? If that's not the case then ok I guess, I just don't understand what we gain by being so restrictive.  But I'm good if you guys are good. Hopefully this means getting rid of `emit_debug()` altogether though.

pkumbh631: Sure, we can address this in a new ticket

brendanobra (approved):

tfarz123 (approved):

PR #768, opened 2025-03-27
fix: Tokio thread lock up issue

satlead:
## What

Tokio Threads seems to lock up after some processing 

## Why

Relates to https://github.com/tokio-rs/tokio/pull/7094

## How

Upgrades tokio to a version where this is patched also more logs to capture issues in future.

## Test

## Checklist

### DIFF
--- a/core/main/src/firebolt/firebolt_gateway.rs
+++ b/core/main/src/firebolt/firebolt_gateway.rs
@@ -30,14 +30,16 @@ use ripple_sdk::{
         },
         observability::{log_signal::LogSignal, metrics_util::ApiStats},
     },
+    chrono::Utc,
     extn::extn_client_message::ExtnMessage,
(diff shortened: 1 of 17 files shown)


Nv-Delle1416 (approved):

pahearn73 (approved):

core/main/src/firebolt/firebolt_gateway.rs
```diff
@@ -112,7 +112,16 @@ impl FireboltGateway {
             .channels_state
             .get_gateway_receiver()
             .expect("Gateway receiver to be available");
+        let handle = Handle::current().metrics();
+
         while let Some(cmd) = firebolt_gateway_rx.recv().await {
```
brendanobra, line 117: Do we really want this at `info`? seems like debug might be more appropriate

brendanobra (approved):

PR #774, opened 2025-04-01
RPPL-2779: Keyboard.email and keyboard.password API's are not returning valid responses

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/bootstrap/start_fbgateway_step.rs
```diff
@@ -58,7 +58,13 @@ impl FireboltGatewayStep {
         let _ = methods.merge(WifiRPCProvider::provide_with_alias(state.clone()));
         let _ = methods.merge(LifecycleRippleProvider::provide_with_alias(state.clone()));
         let _ = methods.merge(CapRPCProvider::provide_with_alias(state.clone()));
-        let _ = methods.merge(KeyboardRPCProvider::provide_with_alias(state.clone()));
+
```
pahearn73, line 61: Keyboard APIs now support app-passthrough schema tags, this change will only configure the built-in Keyboard provider handling if Explicitly not-included in the manifest. This is to allow rollback by way of a manifest update if problems arise.

tfarz123 (approved):

PR #798, opened 2025-04-26
check transformed request for error and return error if it is

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


maggie98choy:
@pahearn73 Currently we don't deter request that is missing required params in endpoint broker. This changes is required to check transformed request error and send error back to sender.

PR #801, opened 2025-04-30
RPPL-2996: added the log_signal for user grant methods

nnaveen979:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


pahearn73 (commented):
Hey @nnaveen979: From the ticket it sounds like @Nv-Delle1416 is asking for:

1. Firebolt request 
2. Process handling 
3. Updating back end services
4. Firebolt response

...I think we already get #1 and #4 for firebolt calls in general, @Nv-Delle1416 correct me if you're asking for more. But for #2 and #3 I think Nik is asking for more logging in [GrantState::update_grant_as_per_policy](https://github.com/rdkcentral/Ripple/blob/main/core/main/src/service/user_grants.rs#L859), where it both checks for settings but also updates grant entries, sends to cloud, etc. Maybe see how you can instrument that function to provide more feedback from calls to `usergrants.grant/deny/clear/request`.

nnaveen979:
> Hey @nnaveen979: From the ticket it sounds like @Nv-Delle1416 is asking for:
> 
> 1. Firebolt request
> 2. Process handling
> 3. Updating back end services
> 4. Firebolt response
> 
> ...I think we already get #1 and #4 for firebolt calls in general, @Nv-Delle1416 correct me if you're asking for more. But for #2 and #3 I think Nik is asking for more logging in [GrantState::update_grant_as_per_policy()](https://github.com/rdkcentral/Ripple/blob/main/core/main/src/service/user_grants.rs#L859), where it both checks for settings but also updates grant entries, sends to cloud, etc. Maybe see how you can instrument that function to provide more feedback from calls to `usergrants.grant/deny/clear/request`.



> Hey @nnaveen979: From the ticket it sounds like @Nv-Delle1416 is asking for:
> 
> 1. Firebolt request
> 2. Process handling
> 3. Updating back end services
> 4. Firebolt response
> 
> ...I think we already get #1 and #4 for firebolt calls in general, @Nv-Delle1416 correct me if you're asking for more. But for #2 and #3 I think Nik is asking for more logging in [GrantState::update_grant_as_per_policy()](https://github.com/rdkcentral/Ripple/blob/main/core/main/src/service/user_grants.rs#L859), where it both checks for settings but also updates grant entries, sends to cloud, etc. Maybe see how you can instrument that function to provide more feedback from calls to `usergrants.grant/deny/clear/request`.

Hi @pahearn73 , i have added log_signals now. please have a look.

core/main/src/firebolt/handlers/user_grants_rpc.rs
```diff
@@ -213,7 +214,13 @@ impl UserGrantsServer for UserGrantsImpl {
         Ok(combined_grant_entries)
     }
 
-    async fn usergrants_grant(&self, _ctx: CallContext, request: GrantRequest) -> RpcResult<()> {
+    async fn usergrants_grant(&self, ctx: CallContext, request: GrantRequest) -> RpcResult<()> {
+        LogSignal::new(
```
pahearn73, line 218: I'm not so sure we need these at the beginning of the handlers, @Nv-Delle1416 don't we already have log signals that cover all firebolt method calls?

core/main/src/firebolt/handlers/user_grants_rpc.rs
```diff
@@ -223,10 +230,23 @@ impl UserGrantsServer for UserGrantsImpl {
         )
         .await;
 
+        LogSignal::new(
+            "usergrants_grant".to_string(),
+            format!("Response:{:?}", result),
```
pahearn73, line 235: The person looking at these logs doesn't know that `Response:{:?}` is the result of a call to `GrantState::update_grant_as_per_policy`. Why not just add `LogSignal` calls inside `GrantState::update_grant_as_per_policy` that logs the return value and why it passed/failed? It's called in 3 different places anyway, so you wouldn't need to do this here (or other places) at all.

nnaveen979: refactoring done, now the log_signal for response used only once. instead of multiple places.

core/main/src/firebolt/handlers/user_grants_rpc.rs
nnaveen979: removed the log_signal's in method entry now.

pahearn73 (approved):

Rajanika (approved):

PR #805, opened 2025-05-09
transition extn client to service

maggie98choy:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/Cargo.toml
+++ b/core/main/Cargo.toml
@@ -43,6 +43,8 @@ contract_tests = [
     "websocket_contract_tests",
 ]
 tdk=[]
+service = []
+
 [dependencies]
 base64.workspace = true
 ripple_sdk = { workspace = true, features = ["full"] }
--- a/core/main/src/bootstrap/boot.rs
+++ b/core/main/src/bootstrap/boot.rs
@@ -60,6 +60,7 @@ pub async fn boot(state: BootstrapState) -> RippleResponse {
     execute_step(StartWsStep, &bootstrap).await?;
     execute_step(StartCommunicationBroker, &bootstrap).await?;
     execute_step(SetupExtnClientStep, &bootstrap).await?;
+    #[cfg(not(feature = "service"))]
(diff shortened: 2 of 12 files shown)


core/main/Cargo.toml
```diff
@@ -43,6 +43,8 @@ contract_tests = [
     "websocket_contract_tests",
 ]
 tdk=[]
+service = []
```
Vinodsathyaseelan, line 46: Please use `ripple_service` instead of the generic `service`

device/thunder_ripple_sdk/src/bootstrap/boot_thunder.rs
```diff
@@ -47,6 +47,7 @@ pub async fn boot_thunder(
     ext_client: ExtnClient,
     _plugin_param: ThunderPluginBootParam,
     device_manifest: &DeviceManifest,
+    local_dev: Option<bool>
```
Vinodsathyaseelan, line 50: Do we need to pass this as a separate param? How is it working in current version with extensions as shared library ?

core/main/src/bootstrap/boot.rs
```diff
@@ -60,7 +60,16 @@ pub async fn boot(state: BootstrapState) -> RippleResponse {
...
     execute_step(SetupExtnClientStep, &bootstrap).await?;
-    execute_step(LoadExtensionsStep, &bootstrap).await?;
+    let load_extensions = std::env::var("RIPPLE_RPC_EXTENSIONS")
+        .ok()
+        .and_then(|s| s.parse::<bool>().ok())
+        .unwrap_or(true);
+    if !load_extensions {
+        debug!("Starting Ripple Service WITHOUT loading extension clients manifest");
+    } else {
+        debug!("Starting Ripple Service with extension clients");
```
Vinodsathyaseelan, line 70: Starting Ripple Services in-proc using shared library.

PR #814, opened 2025-05-12
Update the get_timezone_and_offset function in thunder_device_info.rs according to the latest update of the getTimeZone API

bsenth200:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
```diff
@@ -664,11 +656,32 @@ impl ThunderDeviceInfoRequestProcessor {
     async fn get_all_timezones(
         state: &ThunderState,
     ) -> Result<ThunderAllTimezonesResponse, RippleError> {
+        let params = Some(DeviceChannelParams::Json(
+            json!({
+                "timeZones": [
+                    "America/New_York",
```
satlead, line 662: We need to use the current timezone value we get from the timezone DST as the input for this call and not hardcode this

satlead (changes requested):
Please fix

satlead (approved):

tfarz123 (approved):

PR #815, opened 2025-05-13
feat: support for common thunder_broker subscribe() from Ripple main

nnaveen979:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/core/main/src/broker/broker_utils.rs
+++ b/core/main/src/broker/broker_utils.rs
@@ -17,8 +17,8 @@
 
 use super::{
     endpoint_broker::{
-        BrokerCallback, BrokerCleaner, BrokerConnectRequest, BrokerOutput, BrokerRequest,
-        BrokerSender, BrokerSubMap, EndpointBroker, ATOMIC_ID,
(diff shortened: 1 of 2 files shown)


core/main/src/broker/broker_utils.rs
```diff
@@ -123,4 +159,173 @@ impl BrokerUtils {
...
+        let method = method.clone();
+
+        let broker_sender = BrokerSender {
+            sender: response_tx,
+        };
+        let cleaner = BrokerCleaner {
+            cleaner: Some(c_tx.clone()),
+        };
+
+        let thunder_broker = ThunderBroker::new(broker_sender, subscription_map, cleaner, callback);
```
Vinodsathyaseelan, line 225: Will this result in creating a new instance of ThunderBroker? There should only be a single instance of ThunderBroker in RippleMain. We need to use that existing instance to send the subscription request with a custom callback for receiving event notifications.

nnaveen979:
these changes are not needed actual changes are available in below mentioned PR's.                                                          
Ripple : https://github.com/rdkcentral/Ripple/pull/816
eos-ripple: https://github.com/comcast-firebolt/eos-ripple/pull/843

PR #818, opened 2025-05-19
Add open rpc and JQ rule for Commerce SDK

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/broker/endpoint_broker.rs
```diff
@@ -1140,27 +1140,29 @@ pub trait EndpointBroker {
 
     /// Adds BrokerContext to a given request used by the Broker Implementations
     /// just before sending the data through the protocol
-    fn update_request(rpc_request: &BrokerRequest) -> Result<String, RippleError> {
-        let v = Self::apply_request_rule(rpc_request)?;
+    fn update_request(broker_request: &BrokerRequest) -> Result<String, RippleError> {
```
pahearn73, line 1143: Unrelated, just renamed function variables to properly reflect what they are.

core/main/src/broker/endpoint_broker.rs
```diff
@@ -1648,12 +1650,19 @@ impl BrokerOutputForwarder {
         } else {
             None
         };
-        let parse_result = serde_json::from_slice::<Value>(data);
-        debug!("parse result {:?}", parse_result);
-        if parse_result.is_err() {
-            return Err(RippleError::ParseError);
-        }
-        let result = Some(parse_result.unwrap());
+
```
pahearn73, line 1653: Only validate the response body if it exists, empty bodies are valid for both GET and POST methods.

core/main/src/broker/http_broker.rs
```diff
@@ -24,57 +24,80 @@ use ripple_sdk::{
...
-/*
-
-*/
 
 async fn send_http_request(
     client: &Client<HttpConnector>,
-    method: Method,
     uri: &Uri,
-    path: &str,
+    broker_request: BrokerRequest,
```
pahearn73, line 47: This adds POST support to `HttpBroker`, wherein if the rule has a request transform it is assumed to be the POST request body.

core/main/src/broker/http_broker.rs
```diff
@@ -128,45 +152,49 @@ impl EndpointBroker for HttpBroker {
...
+                            request.rpc.ctx.clone(),
+                        )
+                        .emit_debug();
+
                         let (parts, body) = response.into_parts();
                         let body = body_to_bytes(body).await;
-                        let mut request = request;
-                        if let Ok(json_str) = serde_json::from_slice::<serde_json::Value>(&body).map(|v| vec![v])
+
+                        if body.len() > 0 {
```
pahearn73, line 173: Since we now support POST, only validate response body if it exists.

core/main/src/broker/http_broker.rs
```diff
@@ -374,75 +390,6 @@ mod tests {
         assert!(broker.get_cleaner().cleaner.is_none());
     }
 
-    #[tokio::test]
```
pahearn73, line 377: Added new tests below.

Cargo.lock
```diff
@@ -103,6 +103,25 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
+[[package]]
+name = "ascii-canvas"
```
Vinodsathyaseelan, line 107: We need to investigate why all these packages are being updated. It appears that we are fetching the latest versions of all dependencies from `crates.io` each time.

core/main/src/broker/http_broker.rs
```diff
@@ -24,57 +24,80 @@ use ripple_sdk::{
...
+        .rule
+        .transform
+        .get_transform_data(RuleTransformType::Request)
+    {
+        method = Method::POST;
+
+        let transform_params =
+            match serde_json::from_str::<Vec<Value>>(&broker_request.rpc.params_json) {
+                Ok(mut params) => {
+                    if params.len() > 1 {
```
Vinodsathyaseelan, line 65: Are the statements on lines 65–68 functionally the same as using

`params.pop().unwrap_or(Value::Null)`

?

pahearn73: Yes that's cleaner thanks.

pahearn73: Also updated places where I copied that code from.

Vinodsathyaseelan (dismissed):

tfarz123 (approved):

tfarz123 (approved):

PR #822, opened 2025-05-21
Support for Multiple IPA providers

pahearn73:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/ripple
+++ b/ripple
@@ -93,7 +93,7 @@ case ${1} in
     ;;
     "run")
         cargo build --features local_dev
-        DEVICE_HOST=${2} cargo run --features local_dev core/main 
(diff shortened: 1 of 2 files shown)


ripple
```diff
@@ -93,7 +93,7 @@ case ${1} in
     ;;
     "run")
         cargo build --features local_dev
-        DEVICE_HOST=${2} cargo run --features local_dev core/main 
+        DEVICE_HOST=${2} cargo run --features local_dev --bin ripple
```
pahearn73, line 96: Unrelated: `main` branch now has some of the multiple-binary/services stuff so we need to update this script.

Nv-Delle1416 (approved):

PR #825, opened 2025-05-23
Rppl 2979 enh

satlead:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/main/src/service/apps/delegated_launcher_handler.rs
```diff
@@ -489,80 +489,151 @@ impl DelegatedLauncherHandler {
...
+                    &app_id,
+                    OnRestore,
+                    old_state,
+                    new_state,
+                    None,
+                )
+                .await;
             }
-            (Paused, Terminating) => {
+            (_, Terminating) => {
```
Vinodsathyaseelan, line 625: Did we change the state transition logic? 
As per the HLA, transition to `Terminating` is allowed only from `Paused` state.

PR #826, opened 2025-05-28
RPPL-3292: Extn manifest update

joseinweb:
With JQ rules in place, there is no need for extension manifest anymore. Keeping the structure in place until this file can be removed.

## What

Removed the extensions no longer present

## Why

Ripple is crashing in community builds

## How

Removing dependency to invalid libraries

## Test

Make sure ripple runs without issues.

## Checklist

- [ x] I have self-reviewed this PR


satlead (approved):
Looks good

pahearn73 (approved):

PR #847, opened 2025-07-07
RPPL-3298 : fix: error response added for the thunderclient call fn

nnaveen979:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


device/thunder_ripple_sdk/src/client/device_operator.rs
```diff
@@ -174,11 +174,14 @@ impl DeviceResponseMessage {
...
         sub_id: Option<String>,
     ) -> Option<DeviceResponseMessage> {
+        let json_value = serde_json::to_value(json_resp).ok();
         let mut device_response_msg = None;
         if let Some(res) = &json_resp.result {
             device_response_msg = Some(DeviceResponseMessage::new(res.clone(), sub_id));
-        } else if let Some(er) = &json_resp.error {
-            device_response_msg = Some(DeviceResponseMessage::new(er.clone(), sub_id));
+        } else if let Some(_er) = &json_resp.error {
+            if let Some(val) = json_value {
```
pahearn73, line 182: No need to create `json_value` until here so you should just do something like:

`if let Ok(json_value) = serde_json::to_value(json_resp) {`

...but I'm not quite following why in the error case you're setting the `DeviceResponseMessage` equal to the entire `JsonRpcApiResponse`, won't the `DeviceResponseMessage` already be wrapped? In other words why are the error and non-error cases so different? Maybe @Vinodsathyaseelan can answer that?

device/thunder_ripple_sdk/src/client/plugin_manager.rs
```diff
@@ -238,8 +238,16 @@ impl PluginManager {
...
-            .ok();
+            .await;
+
+        if let Ok(response) = resp.as_ref() {
+            if let Some(error) = response.message.get("error") {
+                error!("Failed to subscribe to statechange, error:{:?}", error);
+            } else {
+                info!("Successfully subscribed to statechange");
+            }
+        }
```
pahearn73, line 249: I think you want an `error!` log here to in case `resp.as_ref()` == `Err(e)`

nnaveen979: added this condition

device/thunder_ripple_sdk/src/processors/thunder_device_info.rs
```diff
@@ -309,13 +330,13 @@ impl ThunderDeviceInfoRequestProcessor {
...
             .await;
-        info!("{}", response.message);
+
         if !check_thunder_response_success(&response) {
             error!("{}", response.message);
             return HashMap::new();
         }
 
-        get_audio_profile_from_value(response.message)
+        get_audio_profile_from_value(response.message.clone())
```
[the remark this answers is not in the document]
nnaveen979: removed the clone().

PR #853, opened 2025-07-17
Memory fix rpc broker

kvfasil:
## What

to reduce heap allocation footprint

## Why

Ripple using more memory in runtime.
## How

move rule engine into share reference.
## Test

heaptrack and runtime memory usage.
## Checklist


core/sdk/src/utils/test_utils.rs
```diff
@@ -29,3 +30,26 @@ where
         panic!("Test failed for ExtnRequest variant: {:?}", request);
     }
 }
+
+pub fn log_memory_usage(label: &str) {
```
Vinodsathyaseelan, line 34: We don't want to use this utility in production code. Please move this under a feature flag and enable that flag only for performance test purpose. 

CC @brendanobra

core/main/src/broker/endpoint_broker.rs
```diff
@@ -494,15 +494,15 @@ impl EndpointBrokerState {
         state
     }
     pub fn with_rules_engine(mut self, rule_engine: RuleEngine) -> Self {
-        self.rule_engine = rule_engine;
+        self.rule_engine = Arc::new(RwLock::new(rule_engine));
```
Vinodsathyaseelan, line 497: Use reference counted rule_engine here.

PR #857, opened 2025-07-18
Memory fix for endpoint broker and manifests.

kvfasil:
## What
to reduce heap allocation footprint

## Why

Ripple using more memory in runtime.

## How

moved rule engine, manifest into share reference.

## Test

heaptrack and runtime memory usage.

## Checklist


core/main/src/broker/endpoint_broker.rs
```diff
@@ -494,15 +494,15 @@ impl EndpointBrokerState {
         state
     }
     pub fn with_rules_engine(mut self, rule_engine: RuleEngine) -> Self {
```
Vinodsathyaseelan, line 496: Us `Arc<RwLock<RuleEngine>>` as the input param

Vinodsathyaseelan (approved):

PR #860, opened 2025-07-22
RPPL-3266: Issues with badger.deviceCapabilities response

nnaveen979:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


pahearn73 (approved):

pahearn73 (approved):

core/main/src/state/cap/generic_cap_state.rs
```diff
@@ -29,14 +29,12 @@ use ripple_sdk::{
...
 };
-use serde::Deserialize;
-use serde_json::json;
 
 use crate::state::platform_state::PlatformState;
 
 #[derive(Clone, Debug, Default)]
 pub struct GenericCapState {
-    supported: Arc<RwLock<HashSet<String>>>,
+    supported: Arc<RwLock<HashSet<FireboltPermission>>>,
```
Vinodsathyaseelan, line 37: Can you explain the changes, reason for the change, impact on other use cases. Not sure how this change is relevant for the original reported problem.

Vinodsathyaseelan (changes requested):
.

core/main/src/state/cap/generic_cap_state.rs
nnaveen979: @Vinodsathyaseelan , in testing i saw the supported_cap is updating the capabilities with extra quotes around the capability string. like "\"xrn:firebolt:capability:protocol:wifi\""  but when we check for "xrn:firebolt:capability:protocol:wifi" availability in supported_cap its failing and that leads to isWifiDevice field false always.  i didn't seen any impact. i can check with QA team if needed.  i have attached the test logs in [RPPL-3266](https://ccp.sys.comcast.net/browse/RPPL-3266).

PR #872, opened 2025-08-07
fix : jq rules files priority issue.

kvfasil:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


pkumbh631 (approved):

core/sdk/src/api/manifest/ripple_manifest_loader.rs
```diff
@@ -397,6 +399,29 @@ impl RippleConfigLoader {
     }
 }
 
+pub fn sort_rules_paths_by_keywords(paths: &mut Vec<String>) {
+    paths.sort_by(|a, b| {
+        // A helper function to determine the priority of a given path.
```
brendanobra, line 404: For later: prefer matcher pattern for this type of thing, is more readable and idiomatic, but get it on next iteration.

PR #878, opened 2025-08-19
feat: Move the openRPC schema validation crate under a feature flag and disable it by default

Vinodsathyaseelan:
## What

1. Made OpenRPC validation support optional in the Ripple repository.
2. Added the openrpc_validation feature flag in Cargo.toml to conditionally include the openrpc_validator dependency.
3. Marked openrpc_validator as an optional dependency, only enabled when the feature is specified.
4. Updated the workspace Cargo.toml to exclude openrpc_validator from the default build.
5. Refactored openrpc_state.rs to use conditional compilation:

When the openrpc_validation feature is enabled, real validation logic is used.
When disabled, dummy types and no-op methods are provided, allowing the code to compile and run without validation.
This change reduces build size and dependencies for default builds, while still allowing full OpenRPC validation when needed.

## Why

Why are these changes needed?

## How

## Test

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,11 @@ members = [
     "core/tdk",
     "device/thunder_ripple_sdk",
     "core/main",
-    "device/mock_device",
-    "openrpc_validator"]
+    "device/mock_device"]
+
+# openrpc_validator is excluded from the default workspace build
+# It will only be included when the openrpc_validation feature is enabled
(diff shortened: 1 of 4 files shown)


brendanobra (dismissed):
Clean and idiomatic :)

Nv-Delle1416 (approved):

PR #887, opened 2025-08-28
fix: if no provider there sending error response.

kvfasil:
## What
There is no provider subscribed we are waiting for a timeout error instead we should be erroring immediately stating no provider.

## Why
As expect immediately respond back as error.

## How
instead of waiting responded back as error.

## Test
integrated players tests

## Checklist

### DIFF
--- a/core/main/src/service/apps/provider_broker.rs
+++ b/core/main/src/service/apps/provider_broker.rs
@@ -287,7 +287,7 @@ impl ProviderBroker {
                 .tx
                 .send(ProviderResponsePayload::GenericError(
                     GenericProviderError {
-                        code: 320001,
+                        code: 32001,
                         message: format!("Provider not found for {}", request.method),
                         data: None,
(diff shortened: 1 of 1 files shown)


core/main/src/service/apps/provider_broker.rs
```diff
@@ -282,8 +282,17 @@ impl ProviderBroker {
                 provider_app_id = Some(provider_method.provider.app_id);
             }
         } else {
-            debug!("queuing provider request");
-            ProviderBroker::queue_provider_request(pst, request);
+            // If no provider found, send error response
+            request
+                .tx
```
brendanobra, line 287: Is it possible to add any context data (provder id, app id, etc) to the returned error message?  Without it, the user is stuck looking at logs i think

brendanobra (approved):

tfarz123 (approved):

PR #891, opened 2025-09-10
RPPL-3225 : added the suggestions field in SearchIntentData struct

nnaveen979:
## What

## Why

Why are these changes needed?

## How

## Test

## Checklist


core/sdk/src/api/device/entertainment_data.rs
```diff
@@ -30,8 +30,7 @@ pub struct ContentProvider {
     pub apis: Vec<String>,
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
-#[cfg_attr(test, derive(PartialEq))]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
```
Vinodsathyaseelan, line 33: Is there any reason for including PartialEq here? I thought this is used only in the test code.

Vinodsathyaseelan (changes requested):
.

Manivannan1802 (approved):

